PR #1, opened 2019-03-09
Allowing e to act as expand for !cases

DarkView:
As stated in commit [ff6812c](https://github.com/Dragory/ZeppelinBot/commit/ff6812cc4e365089fb63af93e1b32a356bf9fb19)
I think allowing e to expand would speed things up a bit. I am also getting annoyed at having to type it out every time


DarkView:
BTW i never used TypeScript so this might contain an syntax error etc. :)

DarkView:
Obsolete as expand has been replaced with --expand (-e)

PR #3, opened 2019-05-20
Lock command V1

DarkView:
Implemented `https://canary.discordapp.com/channels/473085256233123841/534711895803035648/566330108252520448` and removed unused imports
Perm can_lock with only level 100+ being able to lock

Switches between @everyone sendMessage: true and sendMessage: false
Sets to false if permission is inherited


DarkView:
Closing since this is moved off of master

PR #4, opened 2019-05-24
Warning for when someone already has a specified amount of warnings

DarkView:
Also includes the lock command because i suck and forgot to make that its own branch


DarkView:
Probably important to accept his BEFORE #3 since this potentially overrides that (if #3 would also be accepted)

PR #7, opened 2019-05-26
Lock command V1

DarkView:
Cleaned up the mess from yesterday, only one commit now

Functionality still the same as #3


DarkView:
Closed because its not properly done, as they  perms may be on / by default. Might revise at some point

PR #8, opened 2019-05-26
Add a warning when the user has >= specified amount of warns

DarkView:
Same as #4 but clean (as in no commit mess)

### DIFF
--- a/src/plugins/ModActions.ts
+++ b/src/plugins/ModActions.ts
@@ -100,7 +100,7 @@ export class ModActionsPlugin extends ZeppelinPlugin<IModActionsPluginConfig> {
         alert_on_rejoin: false,
         alert_channel: null,
         warn_notify_threshold: 5,
-        warn_threshold_message: "The user already has {priorWarnings} warnings. Proceed anyways?",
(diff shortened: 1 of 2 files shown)


src/plugins/ModActions.ts
```diff
@@ -444,6 +446,14 @@ export class ModActionsPlugin extends ZeppelinPlugin<IModActionsPluginConfig> {
       ),
     );
 
+    const priorWarnAmount = (await casesPlugin.getCaseTypeAmountForUserId(memberToWarn.id, CaseTypes.Warn)) - 1;
+    if (priorWarnAmount >= config.warn_amount_notify_threshhold) {
+      this.sendErrorMessage(
```
Dragory: Perhaps we should ask for verification instead? Kinda like we do above for "Failed to message the user".

DarkView: Sounds like a good idea. Will do on the 13th/14th

DarkView:
Changes made and tested. Sorry for the formatting things, those didnt even show up in my commit preview

DarkView:
Replaced by #36

PR #9, opened 2019-06-05
Get a new VoiceState (Member) so it does not show outdated info

DarkView:
Has been an issue more often than you would think and its really annoying


Dragory:
Does this actually refresh the voice state? Looking at https://discordapp.com/developers/docs/resources/guild#get-guild-member it seems it only returns the regular member info, but no voice state.

DarkView:
At least thats what i do in MRVN, and he always has the most recent one. I remember it not working correctly the other way round. Might be worth testing again.

DarkView:
Obsolete

PR #10, opened 2019-06-17
Include or check to see if the acted on member is the bot

DarkView:


Dragory:
I feel like we should be able to use commands on bots. They can be moved above mods with permission levels anyway if needed.

PR #11, opened 2019-06-17
Massmute command

DarkView:
Also autoformatting and a small change to toString of PluginRuntimeError because it threw compile errors otherwise


DarkView:
These are the same changes i had to make to avoid errors as in #13

DarkView:
Replaced by #35

PR #13, opened 2019-06-25
Fully functional LocatePlugin with alerts in DB and reminders

DarkView:
Basically alerts from MRVN, but this time with duration in nice format (2h30m) and reminders (because you can now potentially make an alert for 1 year and forget it exists, so you get reminded of what it was for.
Also where is there too.
With migrate and everything

### DIFF
--- a/src/PluginRuntimeError.ts
+++ b/src/PluginRuntimeError.ts
@@ -11,11 +11,11 @@ export class PluginRuntimeError {
     this.guildId = guildId;
   }
 
-  [util.inspect.custom](depth, options) {
+  [util.inspect.custom]() {
     return `PRE [${this.pluginName}] [${this.guildId}] ${this.message}`;
   }
 
   toString() {
-    return this[util.inspect.custom]();
+    return this[util.inspect.custom];
   }
 }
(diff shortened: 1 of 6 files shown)


DarkView:
Made two updates to
a) Rename the command to vcalert with vca as short
b) Make vca's list and deletable like reminders
c) Delete all active vca's when a user is banned for said user

src/PluginRuntimeError.ts
```diff
@@ -11,11 +11,11 @@ export class PluginRuntimeError {
...
   }
 
-  [util.inspect.custom](depth, options) {
+  [util.inspect.custom]() {
     return `PRE [${this.pluginName}] [${this.guildId}] ${this.message}`;
   }
 
   toString() {
-    return this[util.inspect.custom]();
+    return this[util.inspect.custom];
```
Dragory: This doesn't seem right. With this change, toString() would return the `util.inspect.custom` *function*, not its result.

Dragory:
Looks good! What are the changes in PluginRuntimeError for, though? Check my comment under "Files changed".

DarkView:
Oh those were not supposed to be in there. Only made to make the compiler shut up because it threw errors otherwise

PR #14, opened 2019-07-27
Fully functioning new starboard system

DarkView:
Loaded from the config


DarkView:
Holding off until i pushed some new changes and figured out an error

PR #37, opened 2020-02-01
Rather temporary exemption implementation

DarkView:
Needs to be changed at some point but should be a quick and dirty fix so we dont prevent too many actual humans from connecting during raids with raidlevel on kick


DarkView:
Obsolete now

PR #41, opened 2020-02-25
Allow softban to message users

DarkView:


Dragory:
Softban is really closer to a kick rather than a ban, so it should probably follow kick messages etc. I'd really love to just delete the command tbh and just have a `-clean` / `-c` flag on `!kick`.

DarkView:
Sounds like a plan, i will look into it!

DarkView:
Replaced by #48

PR #57, opened 2020-05-29
utility: add aliases for clean command

CleverSource:
This pull requests add the aliases "purge" and "prune" to the clean command in the utility plugin.

### DIFF
--- a/backend/src/plugins/Utility.ts
+++ b/backend/src/plugins/Utility.ts
@@ -728,7 +728,7 @@ export class UtilityPlugin extends ZeppelinPlugin<TConfigSchema> {
   }
 
   @d.command("clean", "<count:number>", {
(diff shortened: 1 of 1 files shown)


CleverSource:
> very useful. perhaps `prune` and/or `clear` would be great as well

I'll also add prune since it might be useful for anyone who is most common it.

Dragory:
I feel the word `prune` might cause confusion with actual member pruning. To some degree, I feel this affects `purge` as well (anecdotally, I've seen people call the Discord feature "purging"). `clear` would be a great alias though!

CleverSource:
Hmm, that does make sense. I'll add ``clear`` as an alias later today and remove ``prune`` and ``purge``. Also do you think it would be okay to add ``modlogs`` as an alias to ``cases``?

PR #61, opened 2020-06-21
Role removal mute

roflmaoqwerty:
Added ability to remove and restore roles on mute and unmute. Added these params to Automod and Spam plugins as well


DarkView:
Merged with #140

PR #94, opened 2020-07-27
fix: added ban/unban reasons to their respective api calls

almeidx:


Dragory:
There is an issue here with how Eris handles Audit Log reasons; as far as I'm aware, they need to be manually URL encoded before being passed to the functions.

almeidx:
@Dragory done

Dragory:
Tested and merged! Thank you for the pull request!

PR #95, opened 2020-07-28
[K30] Various Fixes for K30

DarkView:
change -active option on locateuser/follow to be a switch


DarkView:
Add "none" to notify to disable notifications on forcemute, similar to forceban

DarkView:
Implemented the two events to remove mutes on user being banned and re-applying the mute on a member being banned

DarkView:
Previously, this message only displayed when you edited an embed, not when you first posted it

DarkView:
We only aborted on an id < 0, but since the IDs begin at 1 we need to check for <= 0

DarkView:
Same as above, but since we checked against length+1 we crashed "in the opposite direction"

PR #107, opened 2020-08-29
Add Modstats command to display case amount by a given user

DarkView:
`!modstats <moderator:user> [within:delay]`
If within is set, filters to cases between time of command execution and time "within ago"


backend/src/plugins/Slowmode/commands/SlowmodeSetCmd.ts
```diff
@@ -18,7 +18,7 @@ const validModes = ["bot", "native"];
 type TMode = "bot" | "native";
 
 export const SlowmodeSetCmd = slowmodeCmd({
-  trigger: "slowmode",
+  trigger: ["slowmode", "zawarudo"],
```
Dragory, line 21: 👀

PR #109, opened 2020-09-07
List Message Author IDs in Bulk Deletes Logs

greenbigfrog:
This makes searching for deleted messages by a certain User ID a lot easier using discords built in search function


greenbigfrog:
I've updated/fixed the formatting on  this, and actually have tested that this looks like I want it to.

Dragory:
Thank you for the PR! I updated it a bit so the IDs are wrapped in inline code blocks (as they are in other logs).

PR #115, opened 2020-10-06
Allow clean to hard stop at a certain ID

DarkView:
https://canary.discord.com/channels/473085256233123841/534711895803035648/760160134751060028
But slightly modified

Clean <count> amount of messages unless the messages are older than the specified ID

### DIFF
--- a/backend/src/plugins/Utility/commands/CleanCmd.ts
+++ b/backend/src/plugins/Utility/commands/CleanCmd.ts
@@ -8,6 +8,7 @@ import { GuildPluginData } from "knub";
 import { SavedMessage } from "../../../data/entities/SavedMessage";
 import { LogType } from "../../../data/LogType";
 import { allowTimeout } from "../../../RegExpRunner";
+import { snowflakeToTimestamp } from "src/utils/snowflakeToTimestamp";
 
 const MAX_CLEAN_COUNT = 150;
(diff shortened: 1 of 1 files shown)


backend/src/plugins/Utility/commands/CleanCmd.ts
```diff
@@ -8,6 +8,7 @@ import { GuildPluginData } from "knub";
 import { SavedMessage } from "../../../data/entities/SavedMessage";
 import { LogType } from "../../../data/LogType";
 import { allowTimeout } from "../../../RegExpRunner";
+import { snowflakeToTimestamp } from "src/utils/snowflakeToTimestamp";
```
almeidx, line 11: For consistency
```suggestion
import { snowflakeToTimestamp } from "../../../utils/snowflakeToTimestamp";
```

almeidx (changes requested):

backend/src/plugins/Utility/commands/CleanCmd.ts
DarkView: Good change, though that made me notice that we dont need that line at all anymore 😅

almeidx: That was my intention all along ~~(or not)~~

PR #118, opened 2020-10-12
Add moderator to mod action template options

DarkView:


DarkView:
I messed up the branches here, redoing

PR #119, opened 2020-10-12
Add moderator to mod action template options

DarkView:
Passes moderator user object to templateFormatter.

Name: {moderator.username}
ID: {moderator.id}
Discrim: {moderator.discriminator}

The other options technically also work but arent really useful for this


Dragory:
Variables passed to user-controlled templates should _always_ be run through `stripObjectToScalars`!

DarkView:
> Variables passed to user-controlled templates should _always_ be run through `stripObjectToScalars`!

Fixed

PR #125, opened 2020-11-19
Add locks to starboard and include example of level-only starring

DarkView:


backend/src/plugins/Starboard/StarboardPlugin.ts
```diff
@@ -88,6 +88,27 @@ export const StarboardPlugin = zeppelinGuildPlugin<StarboardPluginType>()("starb
                 basic:
                   enabled: true
       ~~~
+
+      ### Limit starboard to a specific level (and above)
+      This is identical to the channel-specific starboard above, but only works for a specific level (>=50).
+      It is noteworthy that this only properly works if the starboard requires just one star.
```
Dragory, line 94: Why is this?

DarkView: Due to a request from a bot users and a few others saying they wanted a level-specific starboard as well
https://canary.discord.com/channels/473085256233123841/739034889399042129/779047601454579744

Dragory: Oh, sorry, I meant this line specifically:
> It is noteworthy that this only properly works if the starboard requires just one star
Doesn't it work with multiple stars too? Just that it doesn't count the stars from those below the specified level.

DarkView: Oh, my bad i misunderstood when the reactions get counted.

PR #131, opened 2020-11-27
Allow automatic tag invoke deletion

DarkView:
Includes config preprocessor to only allow either `delete_with_command` or `xyz_delete_invoke`, never both (Both globally and for every category).
Falls back to `delete_with_command` in case delete_with_command is enabled in an override.

### DIFF
--- a/backend/src/plugins/Tags/types.ts
+++ b/backend/src/plugins/Tags/types.ts
@@ -16,7 +16,7 @@ export const TagCategory = t.type({
   user_category_cooldown: tNullable(t.union([t.string, t.number])), // Per user, per tag category
   global_tag_cooldown: tNullable(t.union([t.string, t.number])), // Any user, per tag
   global_category_cooldown: tNullable(t.union([t.string, t.number])), // Any user, per category
-  category_delete_invoke: tNullable(t.boolean), // Any tag, per tag category
+  auto_delete_command: tNullable(t.boolean), // Any tag, per tag category
 
   tags: t.record(t.string, Tag),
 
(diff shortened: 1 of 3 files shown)


backend/src/plugins/Tags/types.ts
```diff
@@ -16,6 +16,7 @@ export const TagCategory = t.type({
   user_category_cooldown: tNullable(t.union([t.string, t.number])), // Per user, per tag category
   global_tag_cooldown: tNullable(t.union([t.string, t.number])), // Any user, per tag
   global_category_cooldown: tNullable(t.union([t.string, t.number])), // Any user, per category
+  category_delete_invoke: tNullable(t.boolean), // Any tag, per tag category
```
Dragory, line 19: I don't think this needs the `category` prefix, it's only needed for cooldowns as those have a difference betwee per tag and per category.

backend/src/plugins/Tags/types.ts
```diff
@@ -31,6 +32,7 @@ export const ConfigSchema = t.type({
   global_tag_cooldown: tNullable(t.union([t.string, t.number])), // Any user, per tag
   user_cooldown: tNullable(t.union([t.string, t.number])), // Per user
   global_cooldown: tNullable(t.union([t.string, t.number])), // Any tag use
+  global_delete_invoke: t.boolean, // Any tag
```
Dragory, line 35: Same as above, I don't think this needs the `global` prefix.

Dragory:
I think the naming of the configuration option could be changed. Maybe `auto_delete_command` or `delete_invocation`?

DarkView:
I added the global and category prefix to indicate the category option overrides the global one, though i can remove it if you dont think that would be necessary.

PR #139, opened 2020-12-29
Add Tempban capabilities to ban

DarkView:
Behaves the same way temporary mutes would, where time is optional after the user argument


backend/src/migrations/1608753440716-CreateTempBansTable.ts
```diff
@@ -0,0 +1,39 @@
...
+          },
+          {
+            name: "expires_at",
+            type: "datetime",
+          },
+        ],
+      }),
+    );
+  }
+
```
Dragory, line 35: Could add an index on `expires_at` here!

backend/src/plugins/ModActions/ModActionsPlugin.ts
```diff
@@ -165,8 +168,13 @@ export const ModActionsPlugin = zeppelinGuildPlugin<ModActionsPluginType>()("mod
 
     state.mutes = GuildMutes.getGuildInstance(guild.id);
     state.cases = GuildCases.getGuildInstance(guild.id);
+    state.tempbans = GuildTempbans.getGuildInstance(guild.id);
     state.serverLogs = new GuildLogs(guild.id);
 
+    state.unloaded = false;
```
Dragory, line 174: Need to also set this to `true` in `onUnload`

backend/src/plugins/ModActions/commands/BanCmd.ts
```diff
@@ -36,16 +45,82 @@ export const BanCmd = modActionsCmd({
     if (!user.id) {
       return sendErrorMessage(pluginData, msg.channel, `User not found`);
     }
+    const time = args["time"] ? args["time"] : null;
+    console.log(time);
```
Dragory, line 49: Leftover debugging? 👀

backend/src/plugins/ModActions/commands/BanCmd.ts
```diff
@@ -36,16 +45,82 @@ export const BanCmd = modActionsCmd({
...
+        if (!existingTempban && !time) {
+          sendErrorMessage(pluginData, msg.channel, `User is already banned indefinitely.`);
+          return;
+        }
+
+        // Ask the mod if we should update the existing ban
+        const alreadyBannedMsg = await msg.channel.createMessage("User is already banned, update ban?");
+        const reply = await waitForReaction(pluginData.client, alreadyBannedMsg, ["✅", "❌"], msg.author.id);
+
+        alreadyBannedMsg.delete();
```
Dragory, line 81: Could use a `.catch(noop)` (`noop` is from utils I think) so we don't get an error here if someone e.g. deletes the message before reacting.

backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -87,22 +100,31 @@ export async function banUserId(
   // Create a case for this action
   const modId = banOptions.caseArgs?.modId || pluginData.client.user.id;
   const casesPlugin = pluginData.getPlugin(CasesPlugin);
+
+  const noteDetails: string[] = [];
+  const timeUntilUnban = banTime ? humanizeDuration(banTime) : "indefinite";
+  const timeDetails = `Banned ${banTime ? `for ${timeUntilUnban}` : "indefinitely"}`;
+  if (notifyResult.text) noteDetails.push(notifyResult.text);
```
Dragory, line 107: Could do `ucfirst(notifyResult.text)` here like it was previously

dashboard/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "rimraf dist && cross-env NODE_ENV=production webpack --config webpack.config.js",
     "build-debug": "rimraf dist && cross-env NODE_ENV=development webpack --config webpack.config.js",
-    "watch": "cross-env NODE_ENV=development webpack-dev-server",
+    "watch": "cross-env NODE_ENV=development webpack-dev-server --disable-host-check",
```
Dragory, line 9: Probably not needed in this PR

backend/src/plugins/ModActions/commands/BanCmd.ts
DarkView: 300px-Perhaps

backend/src/plugins/ModActions/commands/BanCmd.ts
DarkView: Also added to the previous ban->forceban upgrade

dashboard/package.json
DarkView: Huh, normally i make sure that isnt included in my PRs 🤔

PR #140, opened 2021-01-01
Allow certain or all roles to be removed upon mute and readded on unmute

DarkView:
All credit goes to @roflmaoqwerty, i merely updated it to work with the new version due to time constraints on his side.
Originally PR #61

### DIFF
--- a/backend/src/plugins/Mutes/functions/clearExpiredMutes.ts
+++ b/backend/src/plugins/Mutes/functions/clearExpiredMutes.ts
@@ -17,8 +17,9 @@ export async function clearExpiredMutes(pluginData: GuildPluginData<MutesPluginT
         }
         if (mute.roles_to_restore) {
           const memberOptions: MemberOptions = {};
+          const guildRoles = pluginData.guild.roles;
           memberOptions.roles = Array.from(
-            new Set([...mute.roles_to_restore, ...member.roles.filter(x => x !== muteRole)]),
+            new Set([...mute.roles_to_restore, ...member.roles.filter(x => x !== muteRole && guildRoles.has(x))]),
           );
           member.edit(memberOptions);
         }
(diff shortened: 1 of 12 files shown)


backend/src/plugins/Mutes/functions/clearExpiredMutes.ts
```diff
@@ -14,6 +15,13 @@ export async function clearExpiredMutes(pluginData: GuildPluginData<MutesPluginT
         if (muteRole) {
           await member.removeRole(muteRole);
         }
+        if (mute.roles_to_restore) {
+          const memberOptions: MemberOptions = {};
+          memberOptions.roles = Array.from(
+            new Set([...mute.roles_to_restore, ...member.roles.filter(x => x !== muteRole)]),
+          );
+          member.edit(memberOptions);
```
Dragory, line 23: Might need some validation here to make sure each role we try to restore actually exists

backend/src/plugins/Mutes/functions/unmuteUser.ts
```diff
@@ -36,6 +36,13 @@ export async function unmuteUser(
       if (muteRole && member.roles.includes(muteRole)) {
         await member.removeRole(muteRole);
       }
+      if (existingMute?.roles_to_restore) {
+        const memberOptions: MemberOptions = {};
+        memberOptions.roles = Array.from(
+          new Set([...existingMute.roles_to_restore, ...member.roles.filter(x => x !== muteRole)]),
+        );
+        member.edit(memberOptions);
```
Dragory, line 44: Same validation as above to make sure the roles we're restoring actually exist here

DarkView: Fixed with 499853b

backend/src/plugins/Mutes/functions/clearExpiredMutes.ts
DarkView: Fixed with 499853b

PR #141, opened 2021-01-08
Added aliases for UpdateCmd

yaboyaxis:
Adds `!reason` & `!addnote` to the current `!update` command.


Dragory:
`!addnote` is potentially confusing, as `!note` is a command that is _not_ related to updating an existing case. I think `!reason` would be fine as an alias though!

yaboyaxis:
Makes sense, reflected that change in `a310723`

PR #142, opened 2021-01-12
Disable mentions in logs, allowing mods to be mentioned

DarkView:
Optionally you can return to old behavior by setting allow_user_mentions to true


DarkView:
It is noteworthy that allow_user_mentions is only checked globally because of message chunking preventing per-user control that isnt overly complicated

PR #143, opened 2021-01-12
Add working massunban command

DarkView:
Feedback/Improvement ideas wanted, i feel as if im missing a good way to improve this codes readability and execution time but i cant figure it out right now.


Dragory:
This seems alright for now. Massban is also somewhat slow.

PR #148, opened 2021-01-28
Fix old var names in onMessageCreate

DarkView:


DarkView:
Fixed in ecb9fd43f815bb6b9dcf9553469abd21f610d55f

PR #153, opened 2021-02-06
Add purge alias to !clean/!clear

zeromomentum121:
I keep forgetting that !purge doesn't exist for zepp, so i just ended up making a pr for it

### DIFF
--- a/backend/src/plugins/Utility/commands/CleanCmd.ts
+++ b/backend/src/plugins/Utility/commands/CleanCmd.ts
@@ -50,7 +50,7 @@ async function cleanMessages(
 }
 
 export const CleanCmd = utilityCmd({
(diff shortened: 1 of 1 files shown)


almeidx:
Dragory did not want this back in 2020 https://github.com/Dragory/ZeppelinBot/pull/57#issuecomment-638519946

svcodes:
rip :(

I personally have almost never needed/used the prune feature, and I feel like it would be simpler for a lot of people migrating from other bots, as that's what most bots use (at least from what i've seen)

dexbiobot (changes requested):
Close PR when

zeromomentum121:
i did not realize i could close my own PR 🧠

PR #156, opened 2021-02-12
Allow kicking the user from VC on mute

DarkView:
If any non-id string is entered in `move_to_voice_channel`, the user is kicked from the VC instead of being moved. We do not automatically kick if the option is set to null in order to not make this a breaking change for old, intended behavior

### DIFF
--- a/backend/src/plugins/Mutes/MutesPlugin.ts
+++ b/backend/src/plugins/Mutes/MutesPlugin.ts
@@ -23,6 +23,7 @@ const defaultOptions = {
   config: {
     mute_role: null,
     move_to_voice_channel: null,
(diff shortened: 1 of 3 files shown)


Dragory:
I think having an explicit option might be best here - something like `kick_from_voice: true`

almeidx:
perhaps you could also add an option to voice mute the user

DarkView:
> perhaps you could also add an option to voice mute the user

Since voice mutes are no longer persisted and those would need to be undone in the mutes plugin, im not sure that would work too well



> I think having an explicit option might be best here - something like `kick_from_voice: true`

Can do, ill make Kick take precedent though

PR #157, opened 2021-02-14
Added package-lock.json to .gitignore

Dhruvin-Purohit:
package-lock.json is an automatically generated file, hence not needed.


DarkView:
For your reference, do not remove package-lock.json. It is needed to ensure the correct package versions get installed when using `npm ci`.

Dhruvin-Purohit:
Alr then, closing this pr.

PR #158, opened 2021-02-15
Allow the automod action "reply" to reply in DMs

DarkView:
Also allows the sending of a second message in the server channel to, i.e. let them know they received a DM.
If the DM fails, falls back to in-server reply.
DM messages are not affected by auto_delete, all other "new" messages are.


Dragory:
Will return to this once I've added some safeguards and other things to user DMs

PR #161, opened 2021-03-15
Add options to delete invites from user on kick or ban

DarkView:


backend/src/plugins/ModActions/functions/actualKickMemberCmd.ts
```diff
@@ -77,6 +77,13 @@ export async function actualKickMemberCmd(
     },
   });
 
+  if (pluginData.config.getForMember(msg.member).delete_invites_on_kick) {
+    const userInvites = (await pluginData.guild.getInvites()).filter(inv => inv.inviter?.id === user.id);
+    userInvites.forEach(invite => {
+      invite.delete("User kicked from server");
```
Dragory, line 83: This should either have a perm check above it, a `noop` catch, or a BOT_ALERT if the deletion fails

DarkView: BOT_ALERT added with 6f3aad5

backend/src/plugins/ModActions/functions/actualKickMemberCmd.ts
```diff
@@ -77,6 +77,17 @@ export async function actualKickMemberCmd(
     },
   });
 
+  if (pluginData.config.getForMember(msg.member).delete_invites_on_kick) {
+    const userInvites = (await pluginData.guild.getInvites()).filter(inv => inv.inviter?.id === user.id);
+    userInvites.forEach(invite => {
+      try {
+        invite.delete("User kicked from server");
```
Dragory, line 84: Needs `await` here or the try/catch is skipped

backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -97,6 +97,17 @@ export async function banUserId(
     };
   }
 
+  if (pluginData.config.get().delete_invites_on_ban) {
+    const userInvites = (await pluginData.guild.getInvites()).filter(inv => inv.inviter?.id === user.id);
+    userInvites.forEach(invite => {
+      try {
+        invite.delete("User banned from server");
```
Dragory, line 104: Needs `await` here or the try/catch is skipped

backend/src/plugins/ModActions/functions/actualKickMemberCmd.ts
```diff
@@ -77,6 +77,17 @@ export async function actualKickMemberCmd(
     },
   });
 
+  if (pluginData.config.getForMember(msg.member).delete_invites_on_kick) {
+    const userInvites = (await pluginData.guild.getInvites()).filter(inv => inv.inviter?.id === user.id);
+    userInvites.forEach(invite => {
```
Dragory, line 82: After adding `await` below for `invite.delete()`, it might be best to do a `for (const invite of userInvites)` loop instead and bail from the loop at the first deletion error. The current code would attempt to delete the invites in parallel due to `forEach` which could potentially cause a lot of errors.

backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -97,6 +97,17 @@ export async function banUserId(
     };
   }
 
+  if (pluginData.config.get().delete_invites_on_ban) {
+    const userInvites = (await pluginData.guild.getInvites()).filter(inv => inv.inviter?.id === user.id);
+    userInvites.forEach(invite => {
```
Dragory, line 102: After adding `await` below for `invite.delete()`, it might be best to do a `for (const invite of userInvites)` loop instead and bail from the loop at the first deletion error. The current code would attempt to delete the invites in parallel due to `forEach` which could potentially cause a lot of errors.

PR #164, opened 2021-03-16
Allow users to be DM'd when they use reaction roles

DarkView:


Dragory:
Will return to this once I've added some safeguards and other things to user DMs

PR #166, opened 2021-03-18
Add command to list x recent messages from user

DarkView:
x defaults to 20
-f switch to show full message (trimmed to 50 characters by default)

### DIFF
--- a/backend/src/data/GuildSavedMessages.ts
+++ b/backend/src/data/GuildSavedMessages.ts
@@ -96,6 +96,17 @@ export class GuildSavedMessages extends BaseGuildRepository {
       .getMany();
   }
 
+  getLatestByUser(userId, limit = 20) {
+    return this.messages
+      .createQueryBuilder()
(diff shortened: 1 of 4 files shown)


Dragory:
This definitely needs channel access checks based on the user using the command to not leak e.g. admin channel messages (or other private channels the moderator may not have access to). Ignore me if you already had those and I just missed it!

DarkView:
> This definitely needs channel access checks based on the user using the command to not leak e.g. admin channel messages (or other private channels the moderator may not have access to). Ignore me if you already had those and I just missed it!

Fixed with 025e69b

Dragory:
Hmm, the newest changes would still allow counting how many messages the targeted user has in channels the command user can't read (by comparing the number of shown messages vs. how many were requested). We should either keep requesting messages until we fill the requested amount (or run out of the user's messages in the DB) or, thinking about performance, create a whitelist of channels the command user can see and restrict the messages to those already at query time. There is an index for channel_id, so this should be fairly performant.

PR #167, opened 2021-03-18
Add automod action for setting slowmode

DarkView:


DarkView:
The `channels` array also takes categories, should be included in changelog

Dragory:
> The `channels` array also takes categories

How does this work in practice? From what I can see, the code doesn't resolve child channels for categories, and I don't think categories allow slowmodes directly.

DarkView:
> > The `channels` array also takes categories
> 
> How does this work in practice? From what I can see, the code doesn't resolve child channels for categories, and I don't think categories allow slowmodes directly.

Fixed with d8a8813

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -0,0 +1,52 @@
...
+    duration: "10s",
+  },
+
+  async apply({ pluginData, actionConfig }) {
+    const duration = actionConfig.duration ? convertDelayStringToMS(actionConfig.duration)! : 0;
+
+    for (const channelId of actionConfig.channels) {
+      const channel = pluginData.guild.channels.get(channelId);
+      // 2 = Guild Voice, 5 = Guild News - Both dont allow slowmode
+      if (!channel || channel.type === 2 || channel.type === 5) continue;
```
Dragory, line 23: Stage channels probably don't either. Could filter to allowed channel types rather than skipping specific ones?

Dragory: Could also use `Constants.ChannelTypes` from eris. See: https://github.com/abalabahaha/eris/blob/master/lib/Constants.js#L230

Dragory:
One more nitpick, could use `Constants.ChannelTypes` from eris for clarity. See: https://github.com/abalabahaha/eris/blob/master/lib/Constants.js#L230

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -0,0 +1,53 @@
...
+  },
+
+  async apply({ pluginData, actionConfig }) {
+    const duration = actionConfig.duration ? convertDelayStringToMS(actionConfig.duration)! : 0;
+
+    for (const channelId of actionConfig.channels) {
+      const channel = pluginData.guild.channels.get(channelId);
+      // 0 = Guild Text, 4 = Guild Category - Both dont allow slowmode
+      if (!channel) continue;
+      if (channel.type === 0 || channel.type === 4) continue;
```
Dragory, line 24: This seems to be reversed now, skipping channel types 0 and 4 rather than only allowing them.

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -21,7 +21,7 @@ export const SetSlowmodeAction = automodAction({
       const channel = pluginData.guild.channels.get(channelId);
       // 0 = Guild Text, 4 = Guild Category - Both dont allow slowmode
```
Dragory, line 22: ^ could be clarified to "other channel types don't allow slowmode"

PR #168, opened 2021-03-19
Allow filtering !cases to certain types using switches

DarkView:
It is also possible to reverse the filters using the -r flag.
Examples:

`!cases 108552944961454080 -m -b` Only shows mutes and bans
`!cases 108552944961454080 -r -m` Shows all types *except* for mutes

### DIFF
--- a/backend/src/plugins/ModActions/commands/CasesUserCmd.ts
+++ b/backend/src/plugins/ModActions/commands/CasesUserCmd.ts
@@ -21,12 +21,12 @@ const opts = {
   expand: ct.bool({ option: true, isSwitch: true, shortcut: "e" }),
   hidden: ct.bool({ option: true, isSwitch: true, shortcut: "h" }),
   reverseFilters: ct.switchOption({ shortcut: "r" }),
-  onlyNotes: ct.switchOption({ shortcut: "n" }),
-  onlyWarns: ct.switchOption({ shortcut: "w" }),
-  onlyMutes: ct.switchOption({ shortcut: "m" }),
-  onyUnmutes: ct.switchOption({ shortcut: "um" }),
-  onlyBans: ct.switchOption({ shortcut: "b" }),
-  onyUnbans: ct.switchOption({ shortcut: "ub" }),
+  notes: ct.switchOption({ shortcut: "n" }),
+  warns: ct.switchOption({ shortcut: "w" }),
+  mutes: ct.switchOption({ shortcut: "m" }),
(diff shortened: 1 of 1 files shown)


backend/src/plugins/ModActions/commands/CasesUserCmd.ts
```diff
@@ -15,10 +15,18 @@ import { getGuildPrefix } from "../../../utils/getGuildPrefix";
...
+import { CaseTypes } from "../../../data/CaseTypes";
 
 const opts = {
   expand: ct.bool({ option: true, isSwitch: true, shortcut: "e" }),
   hidden: ct.bool({ option: true, isSwitch: true, shortcut: "h" }),
+  reverseFilters: ct.switchOption({ shortcut: "r" }),
+  onlyNotes: ct.switchOption({ shortcut: "n" }),
+  onlyWarns: ct.switchOption({ shortcut: "w" }),
+  onlyMutes: ct.switchOption({ shortcut: "m" }),
+  onyUnmutes: ct.switchOption({ shortcut: "um" }),
```
Dragory, line 27: Typo ^ `ony` -> `only`

Dragory:
The full names could theoretically also just be `--mutes`, `--warns`, etc. This is how e.g. `pm2` does it: `pm2 logs` lists stderr and stdout, while `pm2 logs --err` only lists errors.

DarkView:
> The full names could theoretically also just be `--mutes`, `--warns`, etc. This is how e.g. `pm2` does it: `pm2 logs` lists stderr and stdout, while `pm2 logs --err` only lists errors.

I was looking for a way of making the function as obvious as possible without having to look at the documentation, but i couldnt come up with anything... I feel like removing the only would make it a little less obvious and people will use the shortcuts and not the full names anyways

DarkView:
> The full names could theoretically also just be `--mutes`, `--warns`, etc. This is how e.g. `pm2` does it: `pm2 logs` lists stderr and stdout, while `pm2 logs --err` only lists errors.

I was looking for a way of making the function as obvious as possible without having to look at the documentation, but i couldnt come up with anything... I feel like removing the only would make it a little less obvious and people will use the shortcuts and not the full names anyways

Dragory:
It's looking great now! Could possibly also add similar support to CasesModCmd at the same time?

DarkView:
In a separate PR probably, yes. For now, this would be good on its own until I'm done with that imo.

PR #171, opened 2021-03-19
chore: improve wording on permissions error for -mod

vcokltfre:
Previously the message was "No permission for -mod" which is a bit ambiguous, and could suggest that the bot doesn't have permission which doesn't make much sense, so I've changed it to "You don't have permission to use -mod" to better convey it's the command author who lacks the required permissions.


vcokltfre:
God damnit how tf is the starboard stuff there...

PR #173, opened 2021-03-19
Add -update/-up argument to automatically update latest/chosen case with !clean

DarkView:


Dragory:
Since this means the Utility plugin depends partially on the ModActions plugin, the plugin's dependencies should be updated accordingly.

PR #178, opened 2021-04-03
AFK Plugin

yaboyaxis:
PR brings AFK functionality, with optional `allow_links` and `allow_invites` options, which allows users to dictate whether they want to allow links/invites in their users statuses.

### DIFF
--- a/backend/src/plugins/AFK/commands/AFKCmd.ts
+++ b/backend/src/plugins/AFK/commands/AFKCmd.ts
@@ -4,40 +4,43 @@ import { sendErrorMessage, sendSuccessMessage } from "../../../pluginUtils";
 import { parseStatusMessage } from "../functions/parseStatusMessage";
 
 export const AfkSetCmd = afkCmd({
-  trigger: ["afk", "afk set"],
-  permission: "can_afk",
+    trigger: ['afk', 'afk set'],
+    permission: 'can_afk',
 
-  signature: {
-    status: ct.string({ rest: true, required: true }),
-  },
+    signature: {
+        status: ct.string({ rest: true, required: true }),
+    },
 
-  async run({ message: msg, args, pluginData }) {
-    // Checks if the user is AFK, if so, return.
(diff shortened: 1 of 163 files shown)


backend/src/plugins/AFK/commands/AFKCmd.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    async run({ message: msg, args, pluginData }) {
+        // Checks if the user is AFK, if so, return.
+        const isAfk = await pluginData.state.afkUsers.getUserAFKStatus(msg.author.id);
+        if (isAfk) return;
+    
+        const status = args.status.join(" ");
+
+        // Check status length
+        if (status.length > 124) {
```
yaboyaxis, line 22: For clarification, I did this in order to prevent people from posting wall-text using their afk statuses.

backend/src/plugins/AFK/events/AFKNotificationEvt.ts
```diff
@@ -0,0 +1,28 @@
+import { sendUserMentionMessage, sendWelcomeBackMessage } from "../functions/buildAFKMessages";
+import { afkEvt } from "../types";
+
+export const AFKNotificationEvt = afkEvt({
+    event: 'messageCreate',
+
+    listener: async ({ pluginData, args: { message } }) => {
+        // Mention Check (if someone mentions the AFK user)
+        if (message.mentions.length) {
+            const afk = await pluginData.state.afkUsers.getUserAFKStatus(message.mentions[0].id);
```
Vetlix, line 10: Might not always be the first mention

yaboyaxis: Yeah, I'll try to make a better system for mention parsing in a bit

yaboyaxis:
Added the `max_status_limit` option, which allows a custom value for how long the users status could be. By default it's 32, but is capped at that to prevent going over the message limit for Discord.

PR #189, opened 2021-04-28
Allow automod to issue tempbans

DarkView:


backend/src/plugins/Automod/actions/ban.ts
```diff
@@ -20,6 +29,7 @@ export const BanAction = automodAction({
 
   async apply({ pluginData, contexts, actionConfig, matchResult }) {
     const reason = actionConfig.reason || "Kicked automatically";
+    const duration = actionConfig.duration ? convertDelayStringToMS(actionConfig.duration)! : undefined;
```
Dragory, line 32: This should already be in milliseconds due to the usage of `tDelayString`

DarkView: tDelayString is only a validator for the format and doesn't convert to ms automatically (this is the same implementation as in mute.ts)

PR #193, opened 2021-05-01
refactor(*): removed unused exception variables

almeidx:
Removed unused exception variables from `try..catch`es


yaboyaxis:
The performance improvement of this change is negligible, but will improve code readability... LGTM

Dragory:
Huh... I didn't know you could leave it out in JS/TS. TIL. Thank you for the PR, and for teaching me something new about JS!

PR #195, opened 2021-05-02
Add various new tag functions

Usoka:
New functions:
`isMention` - ~~Returns the mention if given arg is a valid mention, or empty string otherwise~~ [Updated] Returns boolean value of whether the input is a mention (checks format only, not whether user/role/channel exists).
`setr` - Same as `set` but returns `val`. (defaults to empty string if key is not string).
`lower` - Takes a string and returns it all set to lower case. Returns empty string if arg is not a string.
`upper` - Takes a string and returns it all set to upper case. Returns empty string if arg is not a string.
`upperFirst` - Takes a string and returns the same string with the first character uppercase. Returns empty string if arg is not a string.

### DIFF
--- a/backend/src/plugins/Tags/TagsPlugin.ts
+++ b/backend/src/plugins/Tags/TagsPlugin.ts
@@ -220,10 +220,10 @@ export const TagsPlugin = zeppelinGuildPlugin<TagsPluginType>()("tags", {
 
       isMention: input => {
         if (typeof input !== "string") {
-          return "";
+          return false;
         }
 
-        return input.match(/^<(?:@[!&]?|#)\d+>$/);
+        return /^<(?:@[!&]?|#)\d+>$/.test(input);
       },
     };
 
--- a/backend/src/plugins/Tags/TagsPlugin.ts
+++ b/backend/src/plugins/Tags/TagsPlugin.ts
@@ -203,7 +203,7 @@ export const TagsPlugin = zeppelinGuildPlugin<TagsPluginType>()("tags", {
           return "";
         }
 
-        if (input.match(/^<(?:@(?:!|&)?|#)\d+>$/)) {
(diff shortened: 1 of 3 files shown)


backend/src/plugins/Tags/TagsPlugin.ts
```diff
@@ -217,6 +217,14 @@ export const TagsPlugin = zeppelinGuildPlugin<TagsPluginType>()("tags", {
 
         return "";
       },
+
+      isMention: input => {
+        if (typeof input !== "string") {
+          return "";
+        }
+
+        return input.match(/^<(?:@[!&]?|#)\d+>$/);
```
yaboyaxis, line 226: If you're just expecting a boolean as the return type, [RegExp.test()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) is faster than [String.match()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match).

yaboyaxis (changes requested):

backend/src/plugins/Tags/TagsPlugin.ts
Usoka: Given JS's truthiness with strings I figured it was more versatile for it to return the mention instead of true. That still works as a boolean but also means it can act as a filter (e.g. only keep/return an arg if it is a mention).
Thinking more about this now it would be better to make it a boolean, not as a filter. At least given the naming. I'll change it to that now, thanks!

yaboyaxis (approved):

almeidx (approved):

Dragory:
These will definitely come in handy, thank you! I'll also add `ucfirst` as an alias for `upperFirst`.

PR #196, opened 2021-05-02
added missing duration format to docs

almeidx:
Weeks were missing but [are supported](https://github.com/Dragory/ZeppelinBot/blob/edd78fc9c643a4ba92a21ffc5bd50b8b8e00e448/backend/src/utils.ts#L45)


dexbiobot (approved):

Dragory:
Good catch, thank you!

PR #198, opened 2021-05-05
Added role and emoji info commands

almeidx:
The role-info command was based on the one from https://github.com/Dragory/ZeppelinBot/pull/116

On the emoji-info command, only custom emojis from the guild the command was run on can be used in the command


Dragory:
Thank you for the PR!

PR #199, opened 2021-05-05
Made the nickname argument optional on nickname command

almeidx:
Also added a check to see if the member has a nickname before resetting on the nickname reset command


Dragory:
A common theme with other "setter" commands is that using the command without an argument returns the current value. Perhaps that would work here as well?

So:
`!nickname <userid>` — Shows the current nickname the user has
`!nickname <userid> <new name>` — Changes the nickname (as it is now)
`!nickname reset <userid>` — Resets the nickname (as it is now)

PR #200, opened 2021-05-05
Added can_unban permission for the unban command

almeidx:
The massban and massunban commands have separate permissions, so the ban and unban commands should have different permissions as well


Dragory:
Thank you for the PR! This was definitely an oversight from an earlier time in Zep's development.

PR #205, opened 2021-05-07
Fix issues with both ban commands

DarkView:
- Removes unnecessary check if the user wants to upgrade ban to forceban if target not on server
- Trims down reason to 490 characters. API maximum is 512, and cross-ban bots might add characters onto the reason so I added a small buffer
- Move forceban over to banUserId (as requested in FIXME comment)

### DIFF
--- a/backend/src/plugins/ModActions/functions/banUserId.ts
+++ b/backend/src/plugins/ModActions/functions/banUserId.ts
@@ -87,7 +87,7 @@ export async function banUserId(
     const deleteMessageDays = Math.min(30, Math.max(0, banOptions.deleteMessageDays ?? 1));
     // Trim down reason to 490 words, API limit is 512 and we leave a gap to accomodate cross-ban bots
     if (reason && reason.length >= 490) {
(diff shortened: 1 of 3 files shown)


gdude2002 (approved):

backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -77,6 +77,10 @@ export async function banUserId(
   ignoreEvent(pluginData, IgnoredEventType.Ban, userId);
   try {
     const deleteMessageDays = Math.min(30, Math.max(0, banOptions.deleteMessageDays ?? 1));
+    // Trim down reason to 490 words, API limit is 512 and we leave a gap to accomodate cross-ban bots
+    if (reason && reason.length >= 490) {
+      reason = reason.substring(0, 487) + "...";
```
almeidx, line 82:
```suggestion
      reason = reason.substring(0, 489) + "…";
```

almeidx (changes requested):

PR #207, opened 2021-05-07
docs: Added legacy plugins section

almeidx:
LMK if there are any other legacy plugins


almeidx:
Wrong pull request 😅

Dragory:
Good idea with this one!

PR #216, opened 2021-05-14
Add option for default duration of mutes

Usoka:
Allow users to set a default duration for all mutes (including automod) when a time is not specified. Set via `default_duration: Nullable<tDelayString>` in mutes config. Default/`null` is indefinite mute.

Notes:
- I don't have a full test setup currently. So can't thoroughly test (e.g. mass mute and automod mutes)
- Applies to all mutes. Could be changed to just apply to manual mutes? Though then would be best to move to mod actions.
- If a user tries to mute an already muted user without specifying a length it will update the mute to the default. This mirrors how it currently works with indefinite mutes.

### DIFF
--- a/backend/src/plugins/Mutes/functions/muteUser.ts
+++ b/backend/src/plugins/Mutes/functions/muteUser.ts
@@ -10,6 +10,7 @@ import {
   UserNotificationResult,
   resolveMember,
   UserNotificationMethod,
+  convertDelayStringToMS,
 } from "../../../utils";
 import { renderTemplate } from "../../../templateFormatter";
 import { MemberOptions, TextChannel, User } from "eris";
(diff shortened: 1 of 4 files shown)


backend/src/plugins/Mutes/functions/muteUser.ts
```diff
@@ -245,6 +251,7 @@ export async function muteUser(
 
   return {
     case: theCase,
+    timeUntilUnmute: timeUntilUnmute,
```
almeidx, line 254:
```suggestion
    timeUntilUnmute,
```

almeidx (changes requested):

almeidx (changes requested):

PR #221, opened 2021-05-25
Prevent crash when using !!emoji with no args

Usoka:


almeidx (changes requested):
Set the emoji argument to required instead (`required: true`)
https://github.com/Dragory/ZeppelinBot/blob/0a160218a66a880043bfb74403e7cc25357f521b/backend/src/plugins/Utility/commands/EmojiInfoCmd.ts#L14

almeidx (approved):

PR #224, opened 2021-05-27
Typo correction 

dexbiobot:
Fixed typo where restore_roles_on_mute should have been restore_roles_on_unmute


almeidx:
As is, this is a breaking change. Perhaps adding backwards compatibility would be a better idea

Dragory:
What @almeidx said. This should ideally be accompanied by a config preprocessor that renames the old `restore_roles_on_mute` value to `restore_roles_on_unmute`.

PR #239, opened 2021-07-24
Random Style Improvements

tandpfun:
Heyo!

I've made some random style improvements on the website. I'll write a bit about what I've changed below:

**Guild List Improvements:**
- I cut off long guild names to prevent them from overlapping as seen below:
**Before:**

**After:**

- I added a "Not Allowed" cursor on the disabled buttons:

- I added transition to the config button

**Navbar Improvements**
- Added transitions to the links & "go to dashboard"/"go to docs" buttons
- Added a hover opacity for the dashboard title

**Guild Config Page**
- Added a transition to the save button

**Format Index Files**
The index app file & stuff weren't inside the `npm run format` command's reach so they were never formatted before. I formatted em.

### DIFF
--- a/dashboard/src/components/dashboard/GuildConfigEditor.vue
+++ b/dashboard/src/components/dashboard/GuildConfigEditor.vue
@@ -10,7 +10,7 @@
 
     <div class="flex items-center flex-wrap">
       <h1 class="flex-full md:flex-auto">Config for {{ guild.name }}</h1>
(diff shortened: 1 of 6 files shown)


almeidx:
IMO the Guild Name should be truncated with an ellipsis instead of not rendering the overflow

tandpfun:
Good idea, I'm not the best at css lol and couldn't use the ellipsis without it overlapping with the next guild. I can try to do it today unless you know how.

PR #248, opened 2021-08-14
Change ContextMenu to hardcoded default actions

DarkView:
Sadge

(More complex self-defined stuff coming at some point tho)


almeidx:
image

backend/src/plugins/ContextMenus/actions/mute.ts
```diff
@@ -1,83 +1,61 @@
...
-      if (e instanceof RecoverablePluginError && e.code === ERRORS.NO_MUTE_ROLE_IN_CONFIG) {
-        pluginData.getPlugin(LogsPlugin).log(LogType.BOT_ALERT, {
-          body: `Failed to mute <@!${userId}> in ContextMenu action \`${actionName}\` because a mute role has not been specified in server config`,
-        });
-      } else {
-        throw e;
-      }
+import { ContextMenuPluginType } from "../types";
+
+export async function muteAction(pluginData: GuildPluginData<ContextMenuPluginType>, duration, interaction) {
```
almeidx, line 13: Give types to the duration and interaction

backend/src/plugins/ContextMenus/actions/userInfo.ts
```diff
@@ -0,0 +1,20 @@
+import { GuildPluginData } from "knub";
+import { UtilityPlugin } from "../../../plugins/Utility/UtilityPlugin";
+import { ContextMenuPluginType } from "../types";
+
+export async function userInfoAction(pluginData: GuildPluginData<ContextMenuPluginType>, interaction) {
```
almeidx, line 5: Add a type annotation to the `interaction`

backend/src/plugins/ContextMenus/utils/loadAllCommands.ts
```diff
@@ -1,22 +1,25 @@
...
+  const cfg = pluginData.config.get();
   const newCommands: ApplicationCommandData[] = [];
   const addedNames: string[] = [];
 
-  for (const [name, configAction] of Object.entries(actions)) {
-    if (!configAction.enabled) continue;
+  for (const [name, label] of Object.entries(hardcodedContext)) {
+    if (!cfg[name]) continue;
 
+    const type = name.startsWith("user") ? 2 : 3;
```
almeidx, line 17: Perhaps use the enum from the discord.js Constants called `ApplicationCommandTypes` instead of magic numbers

backend/src/plugins/ModActions/ModActionsPlugin.ts
```diff
@@ -181,6 +182,12 @@ export const ModActionsPlugin = zeppelinGuildPlugin<ModActionsPluginType>()({
       };
     },
 
+    hasMutePermission(pluginData) {
+      return (member: GuildMember, channelId: string) => {
```
almeidx, line 186: This technically doesn't change the type, but why not 

```suggestion
      return (member: GuildMember, channelId: Snowflake) => {
```

backend/src/plugins/Utility/UtilityPlugin.ts
```diff
@@ -162,6 +165,18 @@ export const UtilityPlugin = zeppelinGuildPlugin<UtilityPluginType>()({
         cleanCmd(pluginData, args, msg);
       };
     },
+
+    userInfo(pluginData) {
+      return (userId: string, requestMemberId?: string) => {
```
almeidx, line 170:
```suggestion
      return (userId: Snowflake, requestMemberId?: Snowflake) => {
```

backend/src/plugins/Utility/UtilityPlugin.ts
```diff
@@ -162,6 +165,18 @@ export const UtilityPlugin = zeppelinGuildPlugin<UtilityPluginType>()({
...
+        return getUserInfoEmbed(pluginData, userId, false, requestMemberId);
+      };
+    },
+
+    hasPermission(pluginData) {
+      return (member: GuildMember, channelId: string, permission: string) => {
```
almeidx, line 176:
```suggestion
      return (member: GuildMember, channelId: Snowflake, permission: string) => {
```

backend/src/plugins/ContextMenus/utils/loadAllCommands.ts
DarkView: Sadly whenever I tried that I got a module not found error and when I mirror the enum I get a `xyz is not assignable to zyx` error so I decided not to bother. I'll add comments though

metal0:
While this is a significant improvement over the current functionality of the context menu plugin, maybe adding support for discord's own permissions filtering here would also be nice. Though it wouldn't fit with the current overwrites.

Mainly so that 99% of members in the guild who won't have permission to execute the commands anyway don't even see them in the first place.

https://discord.js.org/#/docs/main/stable/class/ApplicationCommand?scrollTo=permissions

Maybe the `extra` field on config overrides could be used for this?

Dragory:
@metal0 Zeppelin/Knub will eventually try to mirror Discord's permissions directly. This should be more or less possible once channel-based permissions are possible (except for custom override criteria such as antiraid level). Until then, I don't think adding `extra` field criteria separately would be smart, as backwards compatibility would have to be maintained even in the future.

PR #258, opened 2021-08-24
fixed max stickers and premium tier on serverinfocmd

almeidx:


metal0 (approved):

Dragory:
Thank you for the fix!

PR #259, opened 2021-08-25
Automod invite filter updated regex

metal0:
Graphic on discord pointed out this new bypass, just updating the regex.
`https://discord.com/friend-invite/code#`

Also, currently automod will double-trigger invite (and probably others as well) filters due to embeds (message update), even if the violation is the exact same. I'm not sure what the best fix and impl is here, so I won't write up anything, but in theory I imagine it would have to be either checking message content is the exact same on msg update, or just ignoring duplicate violations on the same filter for the same data (i.e. if alrd flagged invite code `1234`, let's not flag it again for the same message)


Dragory:
Thank you for the PR!

PR #262, opened 2021-08-27
Starboard first-pass fix

metal0:
I tested pretty much every use-case for starboard and couldn't get it to error out after just the fix to the message deletion

I don't expect you to merge this, code is not very clean and needs a cleanup script still, just wanted to let you know what the starboard issues were (not that they were very hard to track down but w/e)

@almeidx spare me with the code reviews 🙏 i know this doesnt look good 🙏


Dragory:
Let's hope starboard will behave now! Thank you for the PR!

PR #263, opened 2021-08-27
Persist fix

metal0:


Dragory:
Thank you for the fix! This doesn't fix the underlying problem, but should make it work again for cached members.

PR #264, opened 2021-08-27
Logs - role change compare only role IDs instead of a collection

metal0:


Dragory:
Thank you for the fix!

PR #265, opened 2021-08-27
template fns - add strlen and rounding

metal0:
Not entirely sure about the round function, I'm sure @almeidx will provide some very helpful comments though

### DIFF
--- a/backend/src/templateFormatter.ts
+++ b/backend/src/templateFormatter.ts
@@ -389,7 +389,7 @@ const baseValues = {
     return baseValues.upperFirst(arg);
   },
   strlen(arg) {
-    if (typeof arg !== "string") return arg;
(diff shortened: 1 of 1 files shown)


backend/src/templateFormatter.ts
```diff
@@ -388,6 +388,10 @@ const baseValues = {
   ucfirst(arg) {
     return baseValues.upperFirst(arg);
   },
+  strlen(arg) {
+    if (typeof arg !== "string") return 0;
+    return arg.length;
```
almeidx, line 393:
```suggestion
    return [...arg].length;
```

almeidx (changes requested):

Dragory:
Thank you for the PR!

PR #272, opened 2021-08-29
Added thread_create and thread_delete automod triggers

almeidx:
Example config:

image

```yml
plugins:
  automod: 
    config:
      rules:
        on_thread_create:
          triggers:
            - thread_create: {}
          actions:
            alert:
              channel: "705009450855039042"
              text: "{matchSummary}"
        on_thread_delete: 
          triggers:
            - thread_delete:
                parent: "705009450855039042"
          actions:
            alert:
              channel: "705009450855039042"
              text: "{matchSummary}"
```

### DIFF
--- a/backend/src/data/GuildSavedMessages.ts
+++ b/backend/src/data/GuildSavedMessages.ts
@@ -140,38 +140,10 @@ export class GuildSavedMessages extends BaseGuildRepository<SavedMessage> {
     return data;
   }
 
-<<<<<<< HEAD
   protected async _processEntityFromDB(entity: SavedMessage | undefined) {
     if (entity == null) {
       return entity;
     }
-=======
-  public msgToSavedMessage(message: Message): SavedMessage {
-    const postedAt = moment.utc(message.createdTimestamp, "x").format("YYYY-MM-DD HH:mm:ss");
-
-    return {
-      data: this.msgToSavedMessageData(message),
-      id: message.id,
-      guild_id: (message.channel as GuildChannel).guildId,
-      channel_id: message.channelId,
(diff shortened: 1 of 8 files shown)


metal0 (approved):

backend/src/data/GuildSavedMessages.ts
```diff
@@ -35,7 +35,24 @@ export class GuildSavedMessages extends BaseGuildRepository {
     this.toBePermanent = new Set();
   }
 
-  public msgToSavedMessageData(msg: Message): ISavedMessageData {
+  public static msgToSavedMessage(message: Message): SavedMessage {
```
WeebHiroyuki, line 38: Was there any reason you in particular made this a static function?

WeebHiroyuki (approved):

backend/src/data/GuildSavedMessages.ts
Dragory: I'd rather create a guild-specific instance of the class and put it in Automod's `state`. We can look into extracting `msgToSavedMessage` out of the GuildSavedMessages class later if necessary.

almeidx: I just read the class name. I thought it was GuildSavedMessage the whole time.
Also, I made it static because it's being used [here](https://github.com/almeidx/ZeppelinBot/blob/a3c3a44fa7d08432670cca3e6f3034f9639219bc/backend/src/plugins/Automod/events/runAutomodOnThreadEvents.ts#L25)

almeidx: @Dragory Done

PR #274, opened 2021-08-31
Automod start_thread action

metal0:
Adds the `start_thread` action to automod.

### Sample config:
```yaml
plugins:
  automod:
    config:
      rules:
        test_threads:
          enabled: true
          triggers:
            - match_regex:
                patterns:
                  - "!test_thread"
          actions:
            start_thread:
              auto_archive: 1440
              name: "hi {user.username}"
              private: false
              slowmode: "10s"
              limit_per_channel: 2
            clean: false
```

### Problems:
- Due to the general functionality of automod, these threads generated by the bot cannot be deleted, deleting a thread will cause a message_update event on the originating message (i assume to remove the thread message reference ?), which will prompt the bot to trigger automod again, and re-create the thread again.
This could be fixed by saving the message to a db table but I think this is a bit out-of-bounds of this PR and just something that should be fixed on automod itself (maybe add a `allow_duplication` boolean and store actions taken, if `allow_duplication` is false, don't allow the same action from the same rule to apply again)
- There is no super easy & reliable way to check whose message triggered the new thread retroactively, getting the first cached message in a thread could be impossible after a lot of messages have been sent in that channel, it's something that, again, would have to be saved to a DB table.

### Ideas:
Let me know if any of these are worth implementing I guess
- `per_user_limit` - Limit the amount of threads per user (if per_channel is true, only in that channel (?)) 
This would require storing data for sure, so I didn't want to implement this right away
- `mention` - The bot would send a message in the thread channel pinging the users/roles specified
```yaml
mention:
  users: ["1234", "5678"],
  roles: ["1234", "5678"]
  ```
  
  
  
  btw credits to @almeidx as per usual :joy:

### DIFF
--- a/backend/src/plugins/Automod/AutomodPlugin.ts
+++ b/backend/src/plugins/Automod/AutomodPlugin.ts
@@ -146,6 +146,9 @@ const configPreprocessor: ConfigPreprocessorFn<AutomodPluginType> = options => {
         if (rule["actions"]["log"] == null) {
           rule["actions"]["log"] = true;
         }
+        if (rule["actions"]["clean"] === true && rule["actions"]["start_thread"]) {
+          throw new StrictValidationError([`Cannot have both clean and start_thread at rule '${rule.name}'`]);
+        }
       }
     }
   }
--- a/backend/src/plugins/Automod/AutomodPlugin.ts
+++ b/backend/src/plugins/Automod/AutomodPlugin.ts
@@ -114,6 +114,21 @@ const configPreprocessor: ConfigPreprocessorFn<AutomodPluginType> = options => {
                 ]);
               }
             }
+
+            if (triggerObj[triggerName].match_mime_type) {
+              const white = triggerObj[triggerName].match_mime_type.whitelist_enabled;
+              const black = triggerObj[triggerName].match_mime_type.blacklist_enabled;
+
+              if (white && black) {
+                throw new StrictValidationError([
+                  `Cannot have both blacklist and whitelist enabled at rule <${rule.name}/match_mime_type>`,
+                ]);
+              } else if (!white && !black) {
+                throw new StrictValidationError([
+                  `Must have either blacklist or whitelist enabled at rule <${rule.name}/match_mime_type>`,
+                ]);
+              }
+            }
           }
         }
       }
--- a/backend/src/plugins/Automod/AutomodPlugin.ts
+++ b/backend/src/plugins/Automod/AutomodPlugin.ts
@@ -161,7 +161,7 @@ const configPreprocessor: ConfigPreprocessorFn<AutomodPluginType> = options => {
         if (rule["actions"]["log"] == null) {
           rule["actions"]["log"] = true;
         }
-        if (rule["actions"]["clean"] === true && rule["actions"]["start_thread"]) {
+        if (rule["actions"]["clean"] && rule["actions"]["start_thread"]) {
           throw new StrictValidationError([`Cannot have both clean and start_thread at rule '${rule.name}'`]);
         }
       }
--- a/backend/src/plugins/Automod/actions/startThread.ts
+++ b/backend/src/plugins/Automod/actions/startThread.ts
@@ -31,7 +31,11 @@ export const StartThreadAction = automodAction({
         const threadCount = [
           ...channel.threads.cache
             .filter(
-              tr => tr.ownerId === pluginData.client.application!.id && !tr.archived && tr.parentId === channel.id,
+              tr =>
+                tr.ownerId === pluginData.client.application!.id &&
+                !tr.deleted &&
+                !tr.archived &&
+                tr.parentId === channel.id,
             )
             .keys(),
         ].length; // very short line, yes yes
--- a/backend/src/plugins/Automod/actions/startThread.ts
+++ b/backend/src/plugins/Automod/actions/startThread.ts
@@ -1,9 +1,9 @@
-import { ThreadAutoArchiveDuration } from "discord-api-types";
+import { GuildFeature, ThreadAutoArchiveDuration } from "discord-api-types";
 import { TextChannel } from "discord.js";
 import * as t from "io-ts";
 import { renderTemplate, TemplateSafeValueContainer } from "src/templateFormatter";
 import { convertDelayStringToMS, tDelayString, tNullable } from "src/utils";
-import { userToTemplateSafeUser } from "src/utils/templateSafeObjects";
+import { savedMessageToTemplateSafeSavedMessage, userToTemplateSafeUser } from "src/utils/templateSafeObjects";
 import { noop } from "../../../utils";
 import { automodAction } from "../helpers";
 
@@ -39,12 +39,14 @@ export const StartThreadAction = automodAction({
       }
       return channel.messages.cache.has(c.message.id);
     });
+
+    const guild = await pluginData.guild;
     let autoArchive: ThreadAutoArchiveDuration;
     if (actionConfig.auto_archive === 1440) {
       autoArchive = ThreadAutoArchiveDuration.OneDay;
-    } else if (actionConfig.auto_archive === 4320) {
+    } else if (actionConfig.auto_archive === 4320 && guild.features.includes(GuildFeature.ThreeDayThreadArchive)) {
       autoArchive = ThreadAutoArchiveDuration.ThreeDays;
-    } else if (actionConfig.auto_archive === 10080) {
+    } else if (actionConfig.auto_archive === 10080 && guild.features.includes(GuildFeature.SevenDayThreadArchive)) {
       autoArchive = ThreadAutoArchiveDuration.OneWeek;
     } else {
       autoArchive = ThreadAutoArchiveDuration.OneHour;
@@ -57,6 +59,7 @@ export const StartThreadAction = automodAction({
           str,
           new TemplateSafeValueContainer({
             user: userToTemplateSafeUser(c.user!),
+            msg: savedMessageToTemplateSafeSavedMessage(c.message!),
           }),
         );
       const threadName = await renderThreadName(actionConfig.name ?? "{user.username}#{user.discriminator}s thread");
@@ -64,8 +67,12 @@ export const StartThreadAction = automodAction({
         .create({
           name: threadName,
           autoArchiveDuration: autoArchive,
-          type: actionConfig.private ? "GUILD_PRIVATE_THREAD" : "GUILD_PUBLIC_THREAD",
-          startMessage: !actionConfig.private ? c.message!.id : undefined,
+          type:
+            actionConfig.private && guild.features.includes(GuildFeature.PrivateThreads)
+              ? "GUILD_PRIVATE_THREAD"
+              : "GUILD_PUBLIC_THREAD",
+          startMessage:
+            !actionConfig.private && guild.features.includes(GuildFeature.PrivateThreads) ? c.message!.id : undefined,
         })
         .catch(noop);
       if (actionConfig.slowmode && thread) {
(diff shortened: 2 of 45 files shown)

### CODE REVIEW

File: backend/src/plugins/Automod/actions/startThread.ts [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+import { renderTemplate, TemplateSafeValueContainer } from "src/templateFormatter";
+import { convertDelayStringToMS, tDelayString, tNullable } from "src/utils";
+import { savedMessageToTemplateSafeSavedMessage, userToTemplateSafeUser } from "src/utils/templateSafeObjects";
+import { noop } from "../../../utils";
+import { automodAction } from "../helpers";
+
+export const StartThreadAction = automodAction({
+  configType: t.type({
+    name: tNullable(t.string),
+    auto_archive: tNullable(t.number),
```
Dragory, line 13: Due to the way this is handled, we might want to consider making this a `tDelayString` that's validated to be one of the valid options.

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,90 @@
...
+      const channel = pluginData.guild.channels.cache.get(c.message!.channel_id) as TextChannel;
+      const renderThreadName = async str =>
+        renderTemplate(
+          str,
+          new TemplateSafeValueContainer({
+            user: userToTemplateSafeUser(c.user!),
+            msg: savedMessageToTemplateSafeSavedMessage(c.message!),
+          }),
+        );
+      const threadName = await renderThreadName(actionConfig.name ?? "{user.username}#{user.discriminator}s thread");
```
Dragory, line 69:
```suggestion
      const threadName = await renderThreadName(actionConfig.name ?? "{user.username}#{user.discriminator}'s thread");
```

metal0: I've tested this already. Discord ignores the `

backend/src/plugins/Automod/actions/startThread.ts
metal0: Done.
Not sure if this was the best approach calculations-wise, please check it out.

metal0:
@Dragory Since #272 and #275 exist, should the `slowmode` option from this method be removed now?

backend/src/plugins/Automod/actions/startThread.ts
almeidx: @metal0 Use `user.tag` instead

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+  defaultConfig: {
+    limit_per_channel: 5,
+  },
+
+  async apply({ pluginData, contexts, actionConfig, ruleName }) {
+    // check if the message still exists, we don't want to create threads for deleted messages
+    const threads = contexts.filter(c => {
+      if (!c.message || !c.user) return false;
+      const channel = pluginData.guild.channels.cache.get(c.message.channel_id);
+      if (channel?.type !== "GUILD_TEXT" || !channel.isText()) return false; // for some reason the typing here for channel.type defaults to ThreadChannelTypes (?)
```
almeidx, line 28: Use the ChannelTypeStrings enum instead of a magic string

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+      // check against max threads per channel
+      if (actionConfig.limit_per_channel && actionConfig.limit_per_channel > 0) {
+        const threadCount = [
+          ...channel.threads.cache
+            .filter(
+              tr =>
+                tr.ownerId === pluginData.client.application!.id &&
```
almeidx, line 35: Use `client.user!.id` instead

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+                tr.parentId === channel.id,
+            )
+            .keys(),
+        ].length; // very short line, yes yes
+        if (threadCount >= actionConfig.limit_per_channel) return false;
+      }
+      return channel.messages.cache.has(c.message.id);
+    });
+
+    const guild = await pluginData.guild;
```
almeidx, line 47: Unnecessary await

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+    const archiveSet = actionConfig.auto_archive
+      ? Math.ceil(Math.max(convertDelayStringToMS(actionConfig.auto_archive) ?? 0, 0) / MINUTES)
+      : 1400;
+    let autoArchive: ThreadAutoArchiveDuration;
+    if (archiveSet === 1440) {
```
almeidx, line 52: Make an enum or check if discord.js provides one of their own instead of using magic numbers

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+      autoArchive = ThreadAutoArchiveDuration.ThreeDays;
+    } else if (archiveSet === 10080 && guild.features.includes(GuildFeature.SevenDayThreadArchive)) {
+      autoArchive = ThreadAutoArchiveDuration.OneWeek;
+    } else {
+      autoArchive = ThreadAutoArchiveDuration.OneHour;
+    }
+
+    for (const c of threads) {
+      const channel = pluginData.guild.channels.cache.get(c.message!.channel_id) as TextChannel;
+      const renderThreadName = async str =>
```
almeidx, line 64: Add a type annotation to the `str` parameter

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+      autoArchive = ThreadAutoArchiveDuration.OneDay;
+    } else if (archiveSet === 4320 && guild.features.includes(GuildFeature.ThreeDayThreadArchive)) {
+      autoArchive = ThreadAutoArchiveDuration.ThreeDays;
+    } else if (archiveSet === 10080 && guild.features.includes(GuildFeature.SevenDayThreadArchive)) {
+      autoArchive = ThreadAutoArchiveDuration.OneWeek;
+    } else {
+      autoArchive = ThreadAutoArchiveDuration.OneHour;
+    }
+
+    for (const c of threads) {
```
almeidx, line 62: "c" isn't very descriptive here

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,93 @@
...
+        );
+      const threadName = await renderThreadName(actionConfig.name ?? "{user.tag}s thread");
+      const thread = await channel.threads
+        .create({
+          name: threadName,
+          autoArchiveDuration: autoArchive,
+          type:
+            actionConfig.private && guild.features.includes(GuildFeature.PrivateThreads)
+              ? "GUILD_PRIVATE_THREAD"
+              : "GUILD_PUBLIC_THREAD",
```
almeidx, line 80: Use the `ChannelTypeStrings` enum here. If both of those are not in the enum just add them

almeidx (changes requested):

backend/src/plugins/Automod/AutomodPlugin.ts
```diff
@@ -161,6 +161,9 @@ const configPreprocessor: ConfigPreprocessorFn<AutomodPluginType> = options => {
         if (rule["actions"]["log"] == null) {
           rule["actions"]["log"] = true;
         }
+        if (rule["actions"]["clean"] === true && rule["actions"]["start_thread"]) {
```
almeidx, line 164:
```suggestion
        if (rule["actions"]["clean"] && rule["actions"]["start_thread"]) {
```

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,99 @@
+import { GuildFeature, ThreadAutoArchiveDuration } from "discord-api-types";
+import { TextChannel } from "discord.js";
+import * as t from "io-ts";
+import { renderTemplate, TemplateSafeValueContainer } from "src/templateFormatter";
+import { ChannelTypeStrings } from "src/types";
+import { convertDelayStringToMS, MINUTES, tDelayString, tNullable } from "src/utils";
+import { savedMessageToTemplateSafeSavedMessage, userToTemplateSafeUser } from "src/utils/templateSafeObjects";
```
almeidx, line 7:
```suggestion
import { renderTemplate, TemplateSafeValueContainer } from "../../../templateFormatter";
import { ChannelTypeStrings } from "src/types";
import { convertDelayStringToMS, MINUTES, tDelayString, tNullable } from "../../../utils";
import { savedMessageToTemplateSafeSavedMessage, userToTemplateSafeUser } from "../../../utils/templateSafeObjects";
```

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,99 @@
...
+        ].length; // very short line, yes yes
+        if (threadCount >= actionConfig.limit_per_channel) return false;
+      }
+      return channel.messages.cache.has(c.message.id);
+    });
+
+    const guild = pluginData.guild;
+    const archiveSet = actionConfig.auto_archive
+      ? Math.ceil(Math.max(convertDelayStringToMS(actionConfig.auto_archive) ?? 0, 0) / MINUTES)
+      : 1400;
```
almeidx, line 48: use the enum instead of a magic number

almeidx (changes requested):

backend/src/plugins/Automod/actions/startThread.ts
metal0: But I've always wanted to be a magician!! 😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱😱

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,99 @@
+import { GuildFeature, ThreadAutoArchiveDuration } from "discord-api-types";
+import { TextChannel } from "discord.js";
+import * as t from "io-ts";
+import { renderTemplate, TemplateSafeValueContainer } from "../../../templateFormatter";
+import { ChannelTypeStrings } from "src/types";
```
almeidx, line 5:
```suggestion
import { ChannelTypeStrings } from "../../../types";
```

backend/src/plugins/Automod/actions/startThread.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (actionConfig.limit_per_channel && actionConfig.limit_per_channel > 0) {
+        const threadCount = [
+          ...channel.threads.cache
+            .filter(
+              tr =>
+                tr.ownerId === pluginData.client.user!.id && !tr.deleted && !tr.archived && tr.parentId === channel.id,
+            )
+            .keys(),
+        ].length; // very short line, yes yes
+        if (threadCount >= actionConfig.limit_per_channel) return false;
```
almeidx, line 40:
```suggestion
        const threadCount = channel.threads.cache.filter(
          t => t.ownerId === pluginData.client.user!.id && !t.deleted && !t.archived && t.parentId === channel.id,
        ).size;
        if (threadCount >= actionConfig.limit_per_channel) return false;
```

almeidx (changes requested):

backend/src/plugins/Automod/actions/startThread.ts
metal0: NO!!!

`ERROR: backend/src/plugins/Automod/actions/startThread.ts:33:11 - Shadowed name: 't'`

😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎

almeidx (approved):

PR #275, opened 2021-09-01
Reworked automod "set_slowmode" action

metal0:
Now if `channels` is empty or undefined it will set slowmode on the current channel, in terms of functionality I don't think there's any downgrade or loss, however without proper documentation/awareness I imagine this will create issues down the line of people not knowing how this works, similarly to how many other things on the bot are as of right now.

This is mostly for compatibility with #272

### DIFF
--- a/backend/src/plugins/Automod/actions/setSlowmode.ts
+++ b/backend/src/plugins/Automod/actions/setSlowmode.ts
@@ -28,11 +28,13 @@ export const SetSlowmodeAction = automodAction({
       if (
         !channel ||
         !(
-          channel.type === ChannelTypeStrings.TEXT ||
-          ChannelTypeStrings.CATEGORY ||
-          ChannelTypeStrings.PUBLIC_THREAD ||
-          ChannelTypeStrings.NEWS_THREAD ||
-          ChannelTypeStrings.PRIVATE_THREAD
+          [
+            ChannelTypeStrings.TEXT,
+            ChannelTypeStrings.CATEGORY,
+            ChannelTypeStrings.PUBLIC_THREAD,
+            ChannelTypeStrings.NEWS_THREAD,
+            ChannelTypeStrings.PRIVATE_THREAD,
+          ].includes(channel.type)
         )
       ) {
         continue;
--- a/backend/src/plugins/Automod/actions/setSlowmode.ts
+++ b/backend/src/plugins/Automod/actions/setSlowmode.ts
@@ -1,14 +1,13 @@
-import { Snowflake, TextChannel } from "discord.js";
+import { Snowflake, TextChannel, ThreadChannel } from "discord.js";
 import * as t from "io-ts";
 import { ChannelTypeStrings } from "src/types";
-import { LogType } from "../../../data/LogType";
 import { convertDelayStringToMS, isDiscordAPIError, tDelayString, tNullable } from "../../../utils";
 import { automodAction } from "../helpers";
 import { LogsPlugin } from "../../Logs/LogsPlugin";
 
 export const SetSlowmodeAction = automodAction({
   configType: t.type({
-    channels: t.array(t.string),
+    channels: tNullable(t.array(t.string)),
     duration: tNullable(tDelayString),
   }),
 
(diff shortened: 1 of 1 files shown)


backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -1,33 +1,44 @@
...
   defaultConfig: {
     duration: "10s",
   },
 
-  async apply({ pluginData, actionConfig }) {
+  async apply({ pluginData, actionConfig, contexts }) {
     const slowmodeMs = Math.max(actionConfig.duration ? convertDelayStringToMS(actionConfig.duration)! : 0, 0);
-
-    for (const channelId of actionConfig.channels) {
+    const channels = actionConfig.channels ?? new Array();
```
almeidx, line 20:
```suggestion
    const channels = actionConfig.channels ?? [];
```

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -1,33 +1,44 @@
...
       // Only text channels and text channels within categories support slowmodes
-      if (!channel || !(channel.type === ChannelTypeStrings.TEXT || ChannelTypeStrings.CATEGORY)) {
+      if (
+        !channel ||
+        !(
+          channel.type === ChannelTypeStrings.TEXT ||
+          ChannelTypeStrings.CATEGORY ||
+          ChannelTypeStrings.PUBLIC_THREAD ||
+          ChannelTypeStrings.NEWS_THREAD ||
+          ChannelTypeStrings.PRIVATE_THREAD
```
almeidx, line 35: What you had would only check if the channel type was equal to text

```suggestion
          [
            ChannelTypeStrings.TEXT,
            ChannelTypeStrings.CATEGORY,
            ChannelTypeStrings.PUBLIC_THREAD,
            ChannelTypeStrings.NEWS_THREAD,
            ChannelTypeStrings.PRIVATE_THREAD,
          ].includes(channel.type)
```

almeidx (changes requested):

backend/src/plugins/Automod/actions/setSlowmode.ts
metal0: prettier was complaining about this

backend/src/plugins/Automod/actions/setSlowmode.ts
metal0: oooops

almeidx (approved):

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -1,33 +1,32 @@
...
       const channel = pluginData.guild.channels.cache.get(channelId as Snowflake);
-
       // Only text channels and text channels within categories support slowmodes
-      if (!channel || !(channel.type === ChannelTypeStrings.TEXT || ChannelTypeStrings.CATEGORY)) {
-        continue;
-      }
+      if (!channel?.isText() && channel?.type !== ChannelTypeStrings.CATEGORY) continue;
 
-      const channelsToSlowmode: TextChannel[] = [];
+      const channelsToSlowmode: Array<TextChannel | ThreadChannel> = [];
```
WeebHiroyuki, line 29: For preserving consistency with the rest of the code, do you think it'd be better to use `(TextChannel | ThreadChannel)[]`?

almeidx: TSLint disallows it

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -1,33 +1,32 @@
-import { Snowflake, TextChannel } from "discord.js";
+import { Snowflake, TextChannel, ThreadChannel } from "discord.js";
 import * as t from "io-ts";
 import { ChannelTypeStrings } from "src/types";
```
almeidx, line 3:
```suggestion
import { ChannelTypeStrings } from "../../../types";
```

backend/src/plugins/Automod/actions/setSlowmode.ts
```diff
@@ -1,33 +1,32 @@
...
   defaultConfig: {
     duration: "10s",
   },
 
-  async apply({ pluginData, actionConfig }) {
+  async apply({ pluginData, actionConfig, contexts }) {
     const slowmodeMs = Math.max(actionConfig.duration ? convertDelayStringToMS(actionConfig.duration)! : 0, 0);
-
-    for (const channelId of actionConfig.channels) {
+    const channels: string[] = actionConfig.channels ?? [];
```
almeidx, line 20: The actual type will not change but, since it's already being imported in this file, we might as well use it

```suggestion
    const channels: Snowflake[] = actionConfig.channels ?? [];
```

almeidx (changes requested):

almeidx (approved):

PR #276, opened 2021-09-04
DJS Temporary fix for school hub invites

metal0:
For whatever reason, `invite.channel` is currently `null` for these invites, I assume it's a issue with DJS not parsing the type correctly.
For the time being (and for future-proofing) we should check if `channel` exists.

Invite Sample : `9A55gsSJDZ`

image

image

Issue reported by BanTheNons#8297 on discord (460478012794863637)


Dragory:
Thank you for the quick fix!

PR #279, opened 2021-09-05
feat(mod-actions): require reason

WeebHiroyuki:
As [highly requested](https://discord.com/channels/473085256233123841/534711895803035648/877739373607661578), I've implemented a `require_reason` configuration option for the mod actions plugin. You can configure it by the case type. I haven't gotten around testing it as yet, but if any of you are willing to, I'd appreciate it.

### DIFF
--- a/backend/src/plugins/ModActions/ModActionsPlugin.ts
+++ b/backend/src/plugins/ModActions/ModActionsPlugin.ts
@@ -84,6 +84,7 @@ const defaultOptions = {
     can_deletecase: false,
     can_act_as_other: false,
     create_cases_for_manual_actions: true,
+    require_reason: ["warn"],
   },
   overrides: [
     {
--- a/backend/src/plugins/ModActions/commands/NoteCmd.ts
+++ b/backend/src/plugins/ModActions/commands/NoteCmd.ts
@@ -32,7 +32,12 @@ export const NoteCmd = modActionsCmd({
     }
 
     const userName = user.tag;
-    const reason = formatReasonWithAttachments(args.note, [...msg.attachments.values()]);
+    const config = pluginData.config.get();
+    const reason = formatReasonWithAttachments(args.reason, [...msg.attachments.values()]);
+    if (!reason && config.require_reason.includes("note")) {
+      sendErrorMessage(pluginData, msg.channel, "You must include a reason in your note");
+      return;
+    }
 
     const casesPlugin = pluginData.getPlugin(CasesPlugin);
     const createdCase = await casesPlugin.createCase({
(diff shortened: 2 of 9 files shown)


backend/src/plugins/ModActions/commands/NoteCmd.ts
```diff
@@ -32,7 +32,12 @@ export const NoteCmd = modActionsCmd({
     }
 
     const userName = user.tag;
-    const reason = formatReasonWithAttachments(args.note, [...msg.attachments.values()]);
+    const config = pluginData.config.get();
+    const reason = formatReasonWithAttachments(args.reason, [...msg.attachments.values()]);
```
almeidx, line 36:
```suggestion
    const reason = formatReasonWithAttachments(args.note, [...msg.attachments.values()]);
```

almeidx (changes requested):

WeebHiroyuki:
Shouldn't the note and warn commands have required reasons by default, though? What else would be the purpose of the commands? What's a warning without a reason? What's a note without content?

metal0:
> Shouldn't the note and warn commands have required reasons by default, though? What else would be the purpose of the commands? What's a warning without a reason? What's a note without content?

For notes, yeah, I agree.
For warnings, there might be the very odd case where some moderator of some server wants to give a semi-official verbal warning, i.e.
```
bad person: "racist word"
mod: "stop that!"
mod: "!warn @bad person"
```

idk tbh

almeidx:
I agree that the note and warn commands should require the reason. But imo it shouldn't be configurable because, like you said
> What's a warning without a reason? What's a note without content?

WeebHiroyuki:
> > Shouldn't the note and warn commands have required reasons by default, though? What else would be the purpose of the commands? What's a warning without a reason? What's a note without content?
> 
> For notes, yeah, I agree.
> For warnings, there might be the very odd case where some moderator of some server wants to give a semi-official verbal warning, i.e.
> ```
> bad person: "racist word"
> mod: "stop that!"
> mod: "!warn @bad person"
> ```
> 
> idk tbh

I don't think Zeppelin records the context where a case was issued, so it might be best to require reasons for warnings.

metal0 (approved):

backend/src/plugins/ModActions/ModActionsPlugin.ts
```diff
@@ -84,6 +84,7 @@ const defaultOptions = {
     can_deletecase: false,
     can_act_as_other: false,
     create_cases_for_manual_actions: true,
+    require_reason: ["warn"],
```
almeidx, line 87:
```suggestion
    require_reason: [],
```

almeidx (changes requested):

metal0:
how do i retract my approval

Dragory:
Looks good! Can you add a default value for this option as well?

PR #281, opened 2021-09-06
Docs - Document template functions

metal0:
Documents the template functions.

As requested by @Dragory , this is temporarily associated with the Tags plugin for the purpose of later generating this dynamically.

This could **REALLY** use an expandable, but I have no clue how to do it via markdown.

For any changes regarding formatting please just change it yourself on the PR instead of commenting if you have the time.. thanks ~~looking at you @Almeidx~~ 
image

### DIFF
--- a/backend/src/plugins/Tags/TagsPlugin.ts
+++ b/backend/src/plugins/Tags/TagsPlugin.ts
@@ -8,7 +8,7 @@ import { GuildLogs } from "../../data/GuildLogs";
 import { GuildSavedMessages } from "../../data/GuildSavedMessages";
 import { GuildTags } from "../../data/GuildTags";
 import { mapToPublicFn } from "../../pluginUtils";
-import { convertDelayStringToMS } from "../../utils";
+import { convertDelayStringToMS, trimPluginDescription } from "../../utils";
 import { TimeAndDatePlugin } from "../TimeAndDate/TimeAndDatePlugin";
 import { zeppelinGuildPlugin } from "../ZeppelinPluginBlueprint";
 import { TagCreateCmd } from "./commands/TagCreateCmd";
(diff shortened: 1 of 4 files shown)


backend/src/plugins/Tags/TagsPlugin.ts
```diff
@@ -58,6 +60,19 @@ export const TagsPlugin = zeppelinGuildPlugin<TagsPluginType>()({
   showInDocs: true,
   info: {
     prettyName: "Tags",
+    description: trimPluginDescription(`
+      Tags are a way to store and reuse information.
+    `),
```
almeidx, line 65:
```suggestion
    description: "Tags are a way to store and reuse information.",
```

backend/src/plugins/Tags/docs.ts
```diff
@@ -0,0 +1,17 @@
...
+  return definitions
+    .map(def => {
+      const usage = def.signature ?? `(${def.arguments.join(", ")})`;
+      const exampl = def.examples ? def.examples.map(ex => `> \`{${ex}}\``).join("\n") : null;
+      return trimPluginDescription(`
+      ## ${def.name}
+      **${def.description}**\n
+      __Usage__: \`{${def.name}${usage}}\`\n
+      ${exampl ? `__Examples__:\n${exampl}` : ""}\n\n
+      `);
```
almeidx, line 14:
```suggestion
      const examples = def.examples?.map(ex => `> \`{${ex}}\``).join("\n") ?? null;
      return trimPluginDescription(`
      ## ${def.name}
      **${def.description}**\n
      __Usage__: \`{${def.name}${usage}}\`\n
      ${examples ? `__Examples__:\n${examples}` : ""}\n\n
      `);
```

almeidx (changes requested):

almeidx (approved):

PR #286, opened 2021-09-12
Fix thread channels not working for reminders

DarkView:


backend/src/plugins/Reminders/utils/postDueRemindersLoop.ts
```diff
@@ -13,7 +13,7 @@ export async function postDueRemindersLoop(pluginData: GuildPluginData<Reminders
   const pendingReminders = await pluginData.state.reminders.getDueReminders();
   for (const reminder of pendingReminders) {
     const channel = pluginData.guild.channels.cache.get(reminder.channel_id as Snowflake);
-    if (channel && channel instanceof TextChannel) {
+    if (channel && channel.isText()) {
```
almeidx, line 16:
```suggestion
    if (channel?.isText()) {
```

almeidx (changes requested):

DarkView:
Fixed with c7751a9da1da638aabd478e53c7ee6f4bab6cc7f

PR #288, opened 2021-09-13
Fix Persist assigning deleted roles

metal0:
Fixes the persist plugin attempting to check permissions for and assigning deleted roles.


WeebHiroyuki (approved):

almeidx (approved):

Dragory:
Good catch, thank you for the PR!

PR #291, opened 2021-09-14
Added automod unarchive_thread action

metal0:
# Depends on #292
Adds `lock` property to `archive_thread` action to optionally moderator-archive a thread.

The `unlock` property on the `unarchive_thread` action is something I'm not sure is gonna be used too much in practice, though it does serve a different functional purpose to just forcefully unlocking/unarchiving threads.
But change it to your liking /shrug

image


metal0:
@cla-bot check

PR #292, opened 2021-09-14
Added automod thread_archive and thread_unarchive triggers

metal0:
# Followup to #272 

Config example (with #291 )

```yaml
plugins:
  automod:
    config:
      rules:
        test:
          triggers:
            - thread_archive: {}
          actions:
            unarchive_thread: {}
```


Dragory:
Might need to recreate (or force push to) this PR for GitHub to detect the diff properly. Currently it doesn't realize the earlier threads PR was already merged.

PR #293, opened 2021-09-21
allow running cleancmd on news and thread channels

almeidx:


Dragory:
Thank you for the PR!

PR #294, opened 2021-09-21
perf: rolescmd with -counts option

almeidx:
Current setup:
image

With this pr:
image

***

Why the current implementation is slow:
- It uses [`<Role>#members`](https://github.com/discordjs/discord.js/blob/stable/src/structures/Role.js#L139), which by itself uses [`<GuildMember>#roles#cache`](https://github.com/discordjs/discord.js/blob/stable/src/managers/GuildMemberRoleManager.js#L34)

### DIFF
--- a/backend/src/plugins/Utility/commands/RolesCmd.ts
+++ b/backend/src/plugins/Utility/commands/RolesCmd.ts
@@ -34,17 +34,18 @@ export const RolesCmd = utilityCmd({
     if (args.counts) {
       await refreshMembersIfNeeded(guild);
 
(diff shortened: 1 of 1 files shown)


almeidx:
Alternatively, we could get rid of the map and increment the `_memberCount` property instead but we would have to `@ts-expect-error` on it since the property doesn't actually exist in the typings

Throughts @Dragory? 

image

Dragory:
I would definitely prefer using a Map so we don't pollute the role objects.

almeidx:
The current implementation does both. I'll change it to not pollute the role objects then

Dragory:
Thank you for the PR!

PR #297, opened 2021-09-24
Fixed mute error message double negative

Jernik:
I ran into this while testing. This error message gets thrown when the specified mute role is invalid, or when it is not valid. 
The previous error message of "not invalid" made no sense.


Dragory:
Thank you for the PR!

PR #298, opened 2021-09-25
Set comparison_value db fields to be int(32) instead of int(11)

metal0:
In order to be able to handle bigger counter values


Dragory:
As far as I'm aware, INTs are always 32 bits, with the `11` and `32` referring to:
> The length just specifies how many characters to pad when selecting data with the mysql command line client

Reference: https://stackoverflow.com/a/5634147/316944

metal0:
> As far as I'm aware, MySQL/MariaDB INTs are always 32 bits, with the `11` and `32` referring to:
> 
> > The length just specifies how many characters to pad when selecting data with the mysql command line client
> 
> Reference: https://stackoverflow.com/a/5634147/316944

That may be the case, but the bot fails to add counter triggers of values over a million to the database.
See https://discord.com/channels/473085256233123841/534722948246929429/891120343631138838

Dragory:
I don't think that issue is related to the database field type. The values in question are far, far below any INT limits.

PR #299, opened 2021-09-27
feat: allow emojis to be set by environment variables

vcokltfre:
Adds the ability to set custom emoji as environment variables:
- EMOJI_CHECK
- EMOJI_BAN
- EMOJI_UNBAN
- EMOJI_NOTE
- EMOJI_WARN
- EMOJI_KICK
- EMOJI_MUTE
- EMOJI_UNMUTE
- EMOJI_CASE_DELETED
- EMOJI_SOFTBAN

### DIFF
--- a/backend/src/plugins/Cases/caseIcons.ts
+++ b/backend/src/plugins/Cases/caseIcons.ts
@@ -1,13 +1,13 @@
 import { CaseTypes } from "../../data/CaseTypes";
 
 export const caseIcons: Record<CaseTypes, string> = {
-  [CaseTypes.Ban]: "<:case_ban:742540201443721317>",
-  [CaseTypes.Unban]: "<:case_unban:742540201670475846>",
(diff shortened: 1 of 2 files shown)


almeidx (commented):
The `bot.env.example` file should be updated

WeebHiroyuki (changes requested):
The `errorMessage()` function's emoji should be configurable as well.

vcokltfre:
I feel like I should add a note somewhere saying that you dont actually need the emoji env vars to run the bot, and during development you should probably just remove them, but I'm not sure where to put that. Should I add a note in the `bot.env.example` file or in the readme, or somewhere else?

almeidx:
A comment in the env example file should be sufficient

vcokltfre:
Ah that's out of my abilities so I'll let someone else do that lol

PR #301, opened 2021-10-08
Add check to refreshApiKeyExpiryTime to not refresh long lived keys

Jernik:
This PR is to allow long lived ApiKeys to be used in an automated deployment. This is mostly for self-hosted instances, but if a long lived ApiKey is created in the main host by Dragory, this allows it to be used for more than a day. Tested on the Splitgate Server host

### DIFF
--- a/backend/src/data/ApiLogins.ts
+++ b/backend/src/data/ApiLogins.ts
@@ -94,10 +94,7 @@ export class ApiLogins extends BaseRepository {
 
     const login = await this.apiLogins.createQueryBuilder().where("id = :id", { id: loginId }).getOne();
 
-    if (!login) return;
-
-    if (moment.utc(login.expires_at).isSameOrAfter(updatedTime)) return;
-
+    if (!login || moment.utc(login.expires_at).isSameOrAfter(updatedTime)) return;
     await this.apiLogins.update(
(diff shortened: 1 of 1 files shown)


backend/src/data/ApiLogins.ts
```diff
@@ -90,10 +90,18 @@ export class ApiLogins extends BaseRepository {
...
     if (!loginId || !token) return;
 
+    const updatedTime = moment().utc().add(LOGIN_EXPIRY_TIME, "ms");
+
+    const login = await this.apiLogins.createQueryBuilder().where("id = :id", { id: loginId }).getOne();
+
+    if (!login) return;
+
+    if (moment.utc(login.expires_at).isSameOrAfter(updatedTime)) return;
+
```
almeidx, line 100:
```suggestion
    if (!login || moment.utc(login.expires_at).isSameOrAfter(updatedTime)) return;
```

Jernik:
> Wouldn't it be better to just add on to the existing remaining life?

This is a good solution, if I get some time, I'll do that

Dragory:
Adding an explicit `is_permanent` field in the future is compatible with this PR, so it should be fine merging this.

PR #309, opened 2021-12-10
Automod add changeperms action

metal0:
adds server/role/member perm changing action.

does not REPLACE existing permissions for each object (roles/perm overwrites), but instead changes only the ones specified.

if `channel` is not provided, will default to serverwide perms, if a member's ID is provided in `target`, will silently fail if `channel` is not a valid channel.

There's probably some concern here or there for checking that the permissions we're changing doesn't end up being the same as what's already defined on discord's side, to save a few api calls, but I don't think this is a major issue, and I couldn't think of a super simple, clean and easy way to check this with this code.
@almeidx will probably roast me for it so I thought I'd mention it ahead of time just to mention that I did think of this problem.

samples:

```yaml
plugins:
  automod:
    config:
      rules:
        lockdown:
          triggers:
            - match_regex:
                patterns:
                  - 'lockdown'
          actions:
            reply: "Done lockdown"
            log: true
            change_perms:
              target: "{guild.id}" # the @everyone role
              perms:
                SEND_MESSAGES: false
        mute:
          triggers:
            - match_regex:
                patterns:
                  - 'mute'
          actions:
            reply: "Done mute"
            log: true
            change_perms:
              target: "{user.id}"
              channel: "{msg.channel_id}"
              perms:
                SEND_MESSAGES: false # adds a channel override
        unmute:
          triggers:
            - match_regex:
                patterns:
                  - 'unm'
          actions:
            reply: "Done unmute"
            log: true
            change_perms:
              target: "{user.id}"
              channel: "{msg.channel_id}"
              perms:
                SEND_MESSAGES: null # resets channel override
```

### DIFF
--- a/backend/src/plugins/Automod/actions/changePerms.ts
+++ b/backend/src/plugins/Automod/actions/changePerms.ts
@@ -28,7 +28,7 @@ export const ChangePermsAction = automodAction({
         new TemplateSafeValueContainer({
           user: user ? userToTemplateSafeUser(user) : null,
           guild: guildToTemplateSafeGuild(pluginData.guild),
-          msg: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
+          message: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
         }),
       );
     const renderChannel = async (str: string) =>
@@ -37,7 +37,7 @@ export const ChangePermsAction = automodAction({
         new TemplateSafeValueContainer({
           user: user ? userToTemplateSafeUser(user) : null,
           guild: guildToTemplateSafeGuild(pluginData.guild),
-          msg: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
+          message: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
         }),
       );
     const target = await renderTarget(actionConfig.target);
--- a/backend/src/plugins/Automod/actions/changePerms.ts
+++ b/backend/src/plugins/Automod/actions/changePerms.ts
@@ -80,7 +80,7 @@ export const ChangePermsAction = automodAction({
         await channel.permissionOverwrites.delete(target).catch(noop);
         return;
       }
-      await channel.permissionOverwrites.create(target, newPerms).catch(noop);
+      await channel.permissionOverwrites.edit(target, newPerms).catch(noop);
       return;
     }
 
--- a/backend/src/plugins/Automod/actions/changePerms.ts
+++ b/backend/src/plugins/Automod/actions/changePerms.ts
@@ -3,6 +3,12 @@ import * as t from "io-ts";
 import { automodAction } from "../helpers";
 import { tNullable, isValidSnowflake, tPartialDictionary } from "../../../utils";
 import { noop } from "knub/dist/utils";
+import { renderTemplate, TemplateSafeValueContainer } from "../../../templateFormatter";
+import {
+  guildToTemplateSafeGuild,
+  savedMessageToTemplateSafeSavedMessage,
+  userToTemplateSafeUser,
+} from "../../../utils/templateSafeObjects";
 
 export const ChangePermsAction = automodAction({
   configType: t.type({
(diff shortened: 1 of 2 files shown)


backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+    target: t.string,
+    channel: tNullable(t.string),
+    perms: tPartialDictionary(t.keyof(Permissions.FLAGS), tNullable(t.boolean)),
+  }),
+  defaultConfig: {
+    channel: "",
+    perms: {
+      SEND_MESSAGES: true,
+    },
+  },
```
almeidx, line 24: IMO this shouldn't have any defaults
```suggestion
  defaultConfig: {},
```

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+        str,
+        new TemplateSafeValueContainer({
+          msg: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
+        }),
+      );
+    const target = await renderTarget(actionConfig.target);
+    const channelId = actionConfig.channel ? await renderChannel(actionConfig.channel) : null;
+    const role = await pluginData.guild.roles.fetch(target);
+    if (!role) {
+      const member = await pluginData.guild.members.fetch(target);
```
almeidx, line 49:
```suggestion
      const member = await pluginData.guild.members.fetch(target).catch(noop);
```

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+    const renderChannel = async (str: string) =>
+      renderTemplate(
+        str,
+        new TemplateSafeValueContainer({
+          msg: message ? savedMessageToTemplateSafeSavedMessage(message) : null,
+        }),
+      );
+    const target = await renderTarget(actionConfig.target);
+    const channelId = actionConfig.channel ? await renderChannel(actionConfig.channel) : null;
+    const role = await pluginData.guild.roles.fetch(target);
```
almeidx, line 47:
```suggestion
    const role = pluginData.guild.roles.resolve(target);
```

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (!role) {
+      const member = await pluginData.guild.members.fetch(target);
+      if (!member) return;
+    }
+
+    if (channelId && isValidSnowflake(channelId)) {
+      const channel = await pluginData.guild.channels.fetch(channelId);
```
almeidx, line 54:
```suggestion
      const channel = pluginData.guild.channels.resolve(channelId);
```

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (!channel) return;
+      const overwrite = channel.permissionOverwrites.cache.find((pw) => pw.id === target);
+      const allow = new Permissions(overwrite ? overwrite.allow : "0").serialize();
```
almeidx, line 57:
```suggestion
      const allow = new Permissions(overwrite?.allow ?? 0n).serialize();
```

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,98 @@
...
+      const deny = new Permissions(overwrite ? overwrite.deny : "0").serialize();
```
almeidx, line 58:
```suggestion
      const deny = new Permissions(overwrite?.deny ?? 0n).serialize();
```

almeidx (changes requested):

backend/src/plugins/Automod/actions/changePerms.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (typeof newPerms[key] === "boolean") {
+          hasPerms = true;
+          break;
+        }
+      }
+      if (overwrite && !hasPerms) {
+        await channel.permissionOverwrites.delete(target).catch(noop);
+        return;
+      }
+      await channel.permissionOverwrites.create(target, newPerms).catch(noop);
```
Dragory, line 83: Is there potentially a need to handle updating an existing permission overwrite here?

almeidx: Perhaps the `edit()` method should be used here instead of `create()`.

The difference is:
- `create()`: Creates or replaces if already existent
- `edit()`: Edits or creates if non-existent

Dragory: ...huh. So they're more or less identical? I suppose `edit()` doesn't make an API call if the overwrite already exists?

almeidx: They aren't identical because `create()` **replaces** the old overwrite, so you would lose previously defined permissions

metal0: yeah yeah

metal0:
> Thank you for your pull request and welcome to our community. We could not parse the GitHub identity of the following contributors: **metal**. This is most likely caused by a git client misconfiguration; please make sure to:
> 
> 1. check if your git client is configured with an email to sign commits `git config --list | grep email`
> 2. If not, set it up using `git config --global user.email email@example.com`
> 3. Make sure that the git commit email is configured in your GitHub account settings, see https://github.com/settings/emails

didnt ask + dont care

PR #311, opened 2021-12-23
Update Phisherman endpoint

RedSpid3r:
Switch to using `/v2/domains/check/` instead of `/v2/domains/info` for better performance and reliability.

`/check` endpoint now returns required info by Zeppelin:

```
{
    "classification": "suspicious",
    "verifiedPhish": false
}
```


almeidx:
You'd have to make changes to the consumption of the response as well since it has a different format

PR #320, opened 2022-02-17
fix(tags): embed validation

almeidx:
There was a mismatch of types on zod and io-ts
https://github.com/ZeppelinBot/Zeppelin/blob/feeb0488556f10f6b02311476b32797dba314fa5/backend/src/utils.ts#L330
https://github.com/ZeppelinBot/Zeppelin/blob/feeb0488556f10f6b02311476b32797dba314fa5/backend/src/utils.ts#L269

This makes it fail to send tags with embeds that have `timestamp` configured on them since the embed is actually being validated with both

https://github.com/ZeppelinBot/Zeppelin/blob/feeb0488556f10f6b02311476b32797dba314fa5/backend/src/plugins/Tags/util/renderTagFromString.ts#L35
https://github.com/ZeppelinBot/Zeppelin/blob/feeb0488556f10f6b02311476b32797dba314fa5/backend/src/plugins/Tags/util/onMessageCreate.ts#L89

### DIFF
--- a/backend/src/utils.ts
+++ b/backend/src/utils.ts
@@ -327,7 +327,7 @@ export const zEmbedInput = z.object({
   title: z.string().optional(),
   description: z.string().optional(),
   url: z.string().optional(),
-  timestamp: z.number().optional(),
+  timestamp: z.string().optional(),
   color: z.number().optional(),
 
   footer: z.optional(


metal0 (changes requested):
maybe lets think of a way to only use 1 embed structure at most?

Dragory:
Hmm, should we use number instead of string here for both? I think both d.js and the actual API expect it to be a number.

almeidx:
As [discussed in Discord](https://discord.com/channels/473085256233123841/665148424945139712/966784157827072031) explained in Discord, I believe it should stay as a string

PR #334, opened 2022-04-22
feat: add tag aliases support

Rstar284:
Adds tag alias support to Zeppelin's `Tag` plugin.


Rstar284:
I have no idea how to make `<prefix>tag list-aliases <tag>` work. it just creates a new command.
any help would be appreciated :)

rubyowo:
whaa

PR #346, opened 2022-06-15
Fix "Server voice muted" being true even if self-muted, misleading mods into thinking the user was right-click muted by a moderator

DarkView:
Instead split the options into all four, from mute and deaf to selfMute, selfDeaf, serverMute and serverDeaf.
This should provide a lot more clarity.


hawkeye7662:
Does the info command need information about the user being self muted or defeaned?

PR #348, opened 2022-06-22
feat: Tag list search and improved readability

### LINKED ISSUE #347: Better UI for tag list command
The [tag list](https://github.com/ZeppelinBot/Zeppelin/blob/master/backend/src/plugins/Tags/commands/TagListCmd.ts) command has a very minimal layout simply using a codeblock to display all tag names.

This is incredibly hard to read when there are many tags and makes it difficult to find a tag you are looking for.

Suggestions to improve the tags command:

1. Group tags by starting letter and separate each group into sections
	- Possibly embeds can be used for this and pagination if necessary?
2. Allow searching for tags using partial text and/or fuzzy matching
	- For example, to find tags related to github, you could search `!!tags github` and only tags containing that text would appear in the list

DenverCoder1:
Resolves #347 

Improves the readability of the tag list by grouping tags by their first letter for easy scanning.

Additionally, adds the feature to filter tags based on the letters contained within the tag name, for example, the search "ts" would show results for tags named "ts", "test", "typescript", etc. as the letters appear in the tag name in the given order.

Note: this has only been tested on mock data, not an actual running instance.

Example:

image

image

### DIFF
--- a/backend/src/plugins/Tags/commands/TagListCmd.ts
+++ b/backend/src/plugins/Tags/commands/TagListCmd.ts
@@ -1,3 +1,4 @@
+import { commandTypeHelpers as ct } from "../../../commandTypes";
 import { createChunkedMessage } from "../../../utils";
 import { tagsCmd } from "../types";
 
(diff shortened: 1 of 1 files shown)


Dragory:
@cla-bot check

Dragory:
This looks great, thank you for the PR!

PR #378, opened 2023-01-09
fix: crash when the arr argument for concatArr isn't specified

rubyowo:


rubyowo:
https://github.com/ZeppelinBot/CLA/pull/29

backend/src/templateFormatter.ts
```diff
@@ -346,6 +346,7 @@ const baseValues = {
     return [...args].join("");
   },
   concatArr(arr, separator = "") {
+    if(!arr) return "";
```
almeidx, line 349:
```suggestion
    if (!Array.isArray(arr)) return "";
```

almeidx (changes requested):

backend/src/templateFormatter.ts
rubyowo: thx

backend/src/templateFormatter.ts
```diff
@@ -346,6 +346,7 @@ const baseValues = {
     return [...args].join("");
   },
   concatArr(arr, separator = "") {
+    if(!Array.isArray(arr)) return "";
```
almeidx, line 349:
```suggestion
    if (!Array.isArray(arr)) return "";
```

almeidx (changes requested):

Dragory:
@cla-bot check

PR #388, opened 2023-02-16
fix: twemoji cdn url

rubyowo:


rubyowo:
https://github.com/ZeppelinBot/CLA/pull/29

Dragory:
@cla-bot check

Dragory:
Since we use the base URL from the twemoji package nowadays, this should stay updated along with the package.

PR #393, opened 2023-03-10
Fix: Utility-Clean command increased limit & chunk messages to be deleted

metal0:
Increases the default clean command limit (so that running it multiple times wouldn't be necessary) and makes it so messages are actually deleted in chunks so that the bot doesn't run into trouble (even with the current limit of 150 where-as a single bulk-delete call only allows you to delete 100 messages at most)

If the PR'ed limit needs changes, just request or edit it directly on the PR
image


almeidx (approved):

backend/src/plugins/Utility/commands/CleanCmd.ts
```diff
@@ -33,9 +33,15 @@ export async function cleanMessages(
...
-  channel.bulkDelete(idsToDelete);
-  await pluginData.state.savedMessages.markBulkAsDeleted(idsToDelete);
+  // Actually delete the messages (in chunks of 100)
+
+  const chunks = chunkArray(idsToDelete, 100);
+  await Promise.all(
+    chunks.map(async (chunk) => {
+      await channel.bulkDelete(chunk);
+      await pluginData.state.savedMessages.markBulkAsDeleted(chunk);
+    }),
```
almeidx, line 43: blazing fast promise master race 🏎️🚀
```suggestion
    chunks.map((chunk) => Promise.all([
      channel.bulkDelete(chunk),
      pluginData.state.savedMessages.markBulkAsDeleted(chunk),
    ]),
```
probably works

almeidx (changes requested):

almeidx (approved):

PR #399, opened 2023-04-01
Reminders use timestamps

2srqc:
Might need to reformat !reminders tbh


2srqc:
Signed https://github.com/2srqc/CLA/blob/main/signatures/2srqc.md

PR #403, opened 2023-04-10
Add log_message_id to export cases

iamshoXy:
Recently I noticed that after importing case data the message links are no longer in the embed to jump to a specific case. The reason for this is that **log_message_id** is missing when exporting the case data, which is why all the previous message links were no longer available.

### DIFF
--- a/backend/src/api/guilds/importExport.ts
+++ b/backend/src/api/guilds/importExport.ts
@@ -35,7 +35,7 @@ const caseData = z.object({
   is_hidden: z.boolean(),
   pp_id: z.nullable(z.string()),
   pp_name: z.nullable(z.string()),
-  log_message_id: z.nullable(z.string()),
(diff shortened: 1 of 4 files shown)


Dragory:
@cla-bot check

backend/src/api/guilds/importExport.ts
```diff
@@ -35,7 +35,7 @@ const caseData = z.object({
   is_hidden: z.boolean(),
   pp_id: z.nullable(z.string()),
   pp_name: z.nullable(z.string()),
-
+  log_message_id: z.nullable(z.string()),
```
Dragory, line 38: Might need `.optional()` here to not break existing exports which don't include this field

Dragory (approved):

almeidx (approved):

PR #414, opened 2023-07-03
clean up debug from nginx dockerfile

rubyowo:
i forgor


almeidx (approved):

metal0:
@rubyowo This is arguably better to be present for being able to provide better support for selfhosters, and doesn't cause much damage to established prod bots otherwise

PR #416, opened 2023-07-03
feat: per-server avatars & username system fixes

rubyowo:


rubyowo:
Just noticed that it doesn't update the name for old cases. oh well.

almeidx:
Superseded by #436

PR #420, opened 2023-07-29
feat: default ban reason 

iamshoXy:
This adds a default ban reason, as no DM was sent before if no reason was specified.

The default reason can be changed through the option **default_ban_reason**


backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -42,10 +42,12 @@ export async function banUserId(
     };
   }
 
+  reason = reason || (config.default_ban_reason || "No reason specified");
```
almeidx, line 45:
```suggestion
  reason ||= config.default_ban_reason || "No reason specified";
```

backend/src/plugins/ModActions/types.ts
```diff
@@ -91,36 +92,36 @@ export interface IIgnoredEvent {
 
 export type WarnResult =
   | {
-      status: "failed";
-      error: string;
-    }
+    status: "failed";
+    error: string;
+  }
```
almeidx, line 97: All these changes will likely make prettier fail

almeidx (changes requested):

almeidx:
After putting some more thought into this pull request, why limit this to just ban reasons? Why not add something like a `default_reasons` object and allow setting the default messages for `kick`, `mute`, `ban`, etc.?

iamshoXy:
> After putting some more thought into this pull request, why limit this to just ban reasons? Why not add something like a `default_reasons` object and allow setting the default messages for `kick`, `mute`, `ban`, etc.?

I've added the said object, the only issue atm is that it's not a string in the docs and I don't know why

image

backend/src/plugins/ModActions/types.ts
```diff
@@ -23,6 +23,13 @@ export const ConfigSchema = t.type({
   kick_message: tNullable(t.string),
   ban_message: tNullable(t.string),
   tempban_message: tNullable(t.string),
+  default_reasons: tNullable(
+    t.type({
+      mute: t.string,
+      kick: t.string,
+      ban: t.string,
```
almeidx, line 30: These should be nullable too

backend/src/plugins/ModActions/functions/actualMuteUserCmd.ts
```diff
@@ -42,7 +42,11 @@ export async function actualMuteUserCmd(
   }
 
   const timeUntilUnmute = args.time && humanizeDuration(args.time);
-  const reason = args.reason ? formatReasonWithAttachments(args.reason, [...msg.attachments.values()]) : undefined;
+
+  const config = pluginData.config.get();
+  const reason = args.reason
+    ? formatReasonWithAttachments(args.reason, [...msg.attachments.values()])
+    : config.default_reasons?.mute || "No reason specified";
```
almeidx, line 49: This should limit the reason to 512 characters. But now that I look at the `formatReasonWithAttachments()` function, it also does not seem to be limited. Any chance you can modify that too?

backend/src/plugins/ModActions/functions/actualKickMemberCmd.ts
```diff
@@ -67,7 +67,10 @@ export async function actualKickMemberCmd(
     return;
   }
 
-  const reason = formatReasonWithAttachments(args.reason, msg.attachments);
+  const config = pluginData.config.get();
+  const reason = args.reason
+    ? formatReasonWithAttachments(args.reason, msg.attachments)
+    : config.default_reasons?.kick || "No reason specified";
```
almeidx, line 73: Ditto

backend/src/plugins/ModActions/commands/ForcebanCmd.ts
```diff
@@ -61,7 +61,10 @@ export const ForcebanCmd = modActionsCmd({
       mod = args.mod;
     }
 
-    const reason = formatReasonWithAttachments(args.reason, [...msg.attachments.values()]);
+    const config = pluginData.config.get();
+    const reason = args.reason
+      ? formatReasonWithAttachments(args.reason, [...msg.attachments.values()])
+      : config.default_reasons?.ban || "No reason specified";
```
almeidx, line 67: Ditto

backend/src/plugins/ModActions/functions/banUserId.ts
```diff
@@ -42,10 +41,13 @@ export async function banUserId(
     };
   }
 
+  const config = pluginData.config.get();
+  reason ||= config.default_reasons?.ban || "No reason specified";
```
almeidx, line 45: Ditto

almeidx (changes requested):

backend/src/plugins/ModActions/types.ts
iamshoXy, line 1: Instead of limiting the reason in `formatReasonWithAttachments` it might be better to do it here

backend/src/plugins/ModActions/types.ts
```diff
@@ -25,9 +25,9 @@ export const zModActionsConfig = z.strictObject({
   tempban_message: z.nullable(z.string()),
   default_reasons: z.nullable(
     z.object({
-      mute: z.string(),
-      kick: z.string(),
-      ban: z.string(),
```
iamshoXy: Instead of limiting the reason in `formatReasonWithAttachments` it might be better to do it here

PR #422, opened 2023-08-19
fix: correct embed schemas

brawaru:
Correct embed schemas according to the Discord documentation.

Notable change: fixes ability to set `icon_url` to `author` in embeds.

### DIFF
--- a/backend/src/utils.ts
+++ b/backend/src/utils.ts
@@ -275,6 +275,7 @@ export const tEmbed = t.type({
   video: tNullable(
     t.type({
       url: tNullable(t.string),
+      proxy_url: tNullable(t.string),
       width: tNullable(t.number),
       height: tNullable(t.number),
     }),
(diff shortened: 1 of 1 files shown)


brawaru:
CLA pending here — https://github.com/ZeppelinBot/CLA/pull/35

backend/src/utils.ts
```diff
@@ -298,8 +299,8 @@ export const tEmbed = t.type({
     t.type({
       name: t.string,
       url: tNullable(t.string),
-      width: tNullable(t.number),
-      height: tNullable(t.number),
```
almeidx, line 302: Despite the fact that we should not allow setting these, as they are ignored by Discord's API, we cannot remove them from the schema as that would break existing configurations that are defining these properties. This also applies to the schema below

For clarity, I'm only talking about the removal of `width` and `height`.

almeidx (changes requested):

backend/src/utils.ts
almeidx: Additionally, setting proxy URLs should also not be supported as that will also be ignored by Discord.

PR #423, opened 2023-08-30
Automod action to pause invites

Benricheson101:
Adds an automod action `pause_invites` to pause server invites. Works the same as the Pause Invites button in server settings

CleanShot 2023-08-29 at 23 13 54@2x


Benricheson101:
Created a PR for CLA signature.

backend/src/plugins/Automod/actions/pauseInvites.ts
```diff
@@ -0,0 +1,14 @@
...
+export const PauseInvitesAction = automodAction({
+  configType: t.type({
+    paused: t.boolean,
+  }),
+
+  defaultConfig: {},
+
+  async apply({ pluginData, actionConfig }) {
+    await pluginData.guild.disableInvites(actionConfig.paused);
+  },
```
almeidx, line 13: I feel like with this action name, there is no longer an ambiguity towards what it will do, so this should be enough

```suggestion
  configType: t.boolean,

  defaultConfig: true,

  async apply({ pluginData, actionConfig }) {
    await pluginData.guild.disableInvites(actionConfig);
  },
```

almeidx (approved):

backend/src/plugins/Automod/actions/pauseInvites.ts
Benricheson101: I had this originally, but it seems Zeppelin ignores falsy actions. It was able to pause invites when set to `true`, but the action wouldn't run when `false`

almeidx: Interesting
https://github.com/ZeppelinBot/Zeppelin/blob/5a4e50b19de8f19898826ef2e58c6865c03b90c5/backend/src/plugins/Automod/functions/runAutomod.ts#L133-L135

metal0:
@Dragory https://github.com/ZeppelinBot/CLA/pull/36

PR #424, opened 2023-08-30
Add option to trigger antiraid_level only on change

Benricheson101:
Right now, the automod trigger `antiraid_level` runs any time the antiraid level is updated, even if it's value isn't changing (for example, `high` -> `high`). This adds an option `only_on_change` to disable that behavior, making it only run the rule if the value is actually changing (`off` -> `high`, `low` -> `high`, NOT `high` -> `high`).


almeidx (approved):

almeidx:
@cla-bot check

PR #429, opened 2023-10-09
tags: Add one or two template functions

metal0:
adds like three missing template fns and documents some other ones that were missing from the docs

todo:
- [ ] alias `trim_text` => `get_snowflake`
- [ ] convert number types on `convert_base` to string
- [ ] reconsider how to specify available template function based on plugin type in the docs (currently adds `(<plugin>)`


metal0:
HTTP 429 too many template functions :(

PR #431, opened 2023-11-26
Some fixes to !cases

metal0:
Solves:

- https://linear.app/zeppelinbot/issue/ZDEV-45/revamp-cases-command-from-mod-actions-to-use-embed-content
- https://linear.app/zeppelinbot/issue/ZDEV-46/add-pagination-to-cases-of-members
- Consistent member/mod cases-per-page limit
- Fix inconsistency with `lastCaseNum`  using page boundary limits on last page rather than total cases


backend/src/plugins/ModActions/commands/CasesModCmd.ts
```diff
@@ -50,17 +49,18 @@ export const CasesModCmd = modActionsCmd({
...
-        const lastCaseNum = page * casesPerPage;
+        const lastCaseNum = isLastPage ? totalCases : page * casesPerPage;
         const title = `Most recent cases ${firstCaseNum}-${lastCaseNum} of ${totalCases} by ${modName}`;
 
         const embed = {
           author: {
             name: title,
             icon_url: mod instanceof User ? mod.displayAvatarURL() : undefined,
           },
+          description: lines.join("\n"),
```
almeidx, line 62: We should make sure this is below 4096 characters

backend/src/plugins/ModActions/commands/CasesUserCmd.ts
```diff
@@ -90,49 +92,55 @@ export const CasesUserCmd = modActionsCmd({
...
+              totalPages === 1
+                ? `Cases for ${userName} (${lines.length} total)`
+                : `Most recent cases ${firstCaseNum}-${lastCaseNum} of ${cases.length} for ${userName}`;
+
+            const embed = {
+              author: {
+                name: title,
+                icon_url: user instanceof User ? user.displayAvatarURL() : undefined,
+              },
+              description: lines.join("\n"),
```
almeidx, line 120: Ditto

backend/src/plugins/ModActions/commands/CasesUserCmd.ts
```diff
@@ -90,49 +92,55 @@ export const CasesUserCmd = modActionsCmd({
...
+              fields: [
+                {
+                  name: emptyEmbedValue,
+                  value: trimLines(`Use \`${prefix}case <num>\` to see more information about an individual case`),
```
almeidx, line 124: `trimLines()` usage is not needed

almeidx (changes requested):

backend/src/plugins/ModActions/commands/CasesModCmd.ts
metal0: That would require some major refactoring of the createPaginatedMessage since it's stateless and we just get a fixed # of cases per page currently, would have to offset the next page by X without being able to save this state anywhere.
Probably just lower cases per page, or implement ZDEV-62

PR #432, opened 2023-11-26
automod match_link fixes

metal0:
Fixes ZDEV-64
Fixes ZDEV-66

## Why is this impl so cringe?
~~turns out im lazy. also i didnt want to touch the regex itself because i dont want to deal with potential new bypasses originated due to changing it. and it's also tricky since a non-encoded url can probably have `[]().` and still be valid~~

### DIFF
--- a/backend/src/utils.ts
+++ b/backend/src/utils.ts
@@ -647,7 +647,9 @@ export function getUrlsInString(str: string, onlyUnique = false): MatchedURL[] {
       return urls;
     }
 
-    const hostnameParts = matchUrl.hostname.split(".");
+    const hostname = matchUrl.hostname.endsWith(".") ? matchUrl.hostname.slice(0, -1) : matchUrl.hostname;
+
+    const hostnameParts = hostname.split(".");
     const tld = hostnameParts[hostnameParts.length - 1];
     if (tlds.includes(tld)) {
       urls.push(matchUrl);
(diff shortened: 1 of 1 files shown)


backend/src/utils.ts
```diff
@@ -647,7 +648,16 @@ export function getUrlsInString(str: string, onlyUnique = false): MatchedURL[] {
...
     }
 
-    const hostnameParts = matchUrl.hostname.split(".");
+    let hostname = matchUrl.hostname.toLowerCase();
+
+    if (hostname.length > 3) {
+      while (!hostnameTldRegex.test(hostname.at(-1)!)) {
+        if (!hostname.length) break;
+        hostname = hostname.slice(0, -1);
+      }
```
almeidx, line 657:
```suggestion
      hostname = hostname.replace(/[^a-z]+$/, "");
```

backend/src/utils.ts
```diff
@@ -625,6 +625,7 @@ const plainLinkRegex = /((?!https?:\/\/)\S)+\.\S+/; // anything.anything, withou
 // Both of the above, with precedence on the first one
 const urlRegex = new RegExp(`(${realLinkRegex.source}|${plainLinkRegex.source})`, "g");
 const protocolRegex = /^[a-z]+:\/\//;
+const hostnameTldRegex = /^[a-z]$/;
```
almeidx, line 628:
```suggestion
```

almeidx (changes requested):

almeidx (approved):

PR #469, opened 2024-04-02
🔧 Typo VSCode Container Plugin

zayKenyon:
Accidental typo here, the plugin is called 'Dev Containers' not 'Remote - Containers' which is used in the second option for creating the bot.


Dragory:
Looks like the extension has been renamed since the docs were written. Thanks!

PR #472, opened 2024-05-08
fix: api routes didn't use the `API_PATH_PREFIX` router

rubyowo:


Dragory:
Fixed in https://github.com/ZeppelinBot/Zeppelin/pull/455 (which was merged to the "next" branch)

PR #474, opened 2024-05-12
feat: polls

rubyowo:
update DJS version, add new permissions, and add `match_polls` option

### DIFF
--- a/backend/src/data/GuildSavedMessages.ts
+++ b/backend/src/data/GuildSavedMessages.ts
@@ -119,6 +119,16 @@ export class GuildSavedMessages extends BaseGuildRepository<SavedMessage> {
       }));
     }
 
+    if (msg.poll) {
+      data.poll = {
+        answers: Array.from(msg.poll.answers.values()).map((answer) => ({
+          id: answer.id,
+          text: answer.text,
+        })),
+        question: msg.poll.question,
+      };
+    }
+
     return data;
   }
 
--- a/backend/src/plugins/Automod/functions/matchMultipleTextTypesOnMessage.ts
+++ b/backend/src/plugins/Automod/functions/matchMultipleTextTypesOnMessage.ts
@@ -12,9 +12,17 @@ type TextTriggerWithMultipleMatchTypes = {
   match_usernames: boolean;
   match_nicknames: boolean;
   match_custom_status: boolean;
+  match_polls: boolean;
 };
 
-export type MatchableTextType = "message" | "embed" | "visiblename" | "username" | "nickname" | "customstatus";
+export type MatchableTextType =
+  | "message"
+  | "embed"
+  | "visiblename"
+  | "username"
+  | "nickname"
+  | "customstatus"
+  | "polls";
 
 type YieldedContent = [MatchableTextType, string];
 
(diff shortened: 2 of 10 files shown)


almeidx (changes requested):
There's a crash in discord.js when using polls + partials currently (which zeppelin uses), so let's hold this off for now

backend/src/data/GuildSavedMessages.ts
```diff
@@ -119,6 +119,16 @@ export class GuildSavedMessages extends BaseGuildRepository<SavedMessage> {
       }));
     }
 
+    if (msg.poll) {
+      data.poll = {
+        question: msg.poll.question,
+        answers: Array.from(msg.poll.answers.values()).map((answer) => ({
```
almeidx, line 125:
```suggestion
        answers: msg.poll.answers.map((answer) => ({
```

backend/src/utils/permissionNames.ts
```diff
@@ -48,4 +48,8 @@ export const PERMISSION_NAMES = {
   UseExternalSounds: "Use External Sounds",
   UseSoundboard: "Use Soundboard",
   ViewCreatorMonetizationAnalytics: "View Creator Monetization Analytics",
+  SendPolls: "Allows Sending Polls",
+  CreateEvents: "Allows for creating scheduled events, and editing and deleting those created by the current user.",
+  CreateGuildExpressions:
+    "Allows for creating emojis, stickers, and soundboard sounds, and editing and deleting those created by the current user.",
```
almeidx, line 54: This should match what is shown in the role permissions tab in the discord client
Also, any chance you can sort this object alphabetically?
```suggestion
  SendPolls: "Create Polls",
  CreateEvents: "Create Events",
  CreateGuildExpressions: "Create Expressions",
```

backend/src/plugins/Automod/functions/matchMultipleTextTypesOnMessage.ts
```diff
@@ -59,4 +60,8 @@ export async function* matchMultipleTextTypesOnMessage(
       break;
     }
   }
+
+  if (trigger.match_polls && msg.data.poll) {
+    yield ["poll", JSON.stringify(msg.data.poll)];
```
almeidx, line 65: This will not work for the `answers` prop, since `<Collection>.toJSON()` returns an array with the keys only (for whatever reason...)

You're going to have to construct an object with the desireable keys instead, and manually call `answers.map(...)` or `Object.fromEntries(answers.entries())`, depending on the format you want, and then call JSON.stringify on that.

almeidx (changes requested):

backend/src/plugins/Automod/functions/matchMultipleTextTypesOnMessage.ts
rubyowo: The hell...? Thanks for that though I'll fix it

almeidx: Nevermid, this is a SavedMessage so the `answers` prop is an array, not a Collection

PR #477, opened 2024-05-29
fix: dont show hidden cases without -hidden arg

rubyowo:


almeidx (changes requested):
I believe all the uses of `cases.length` below line 110 should be `casesToDisplay.length`

almeidx (approved):

PR #478, opened 2024-05-31
Make getInviteCodesInString check codes with only 8 chars

StanByes:
When we're using the match_invite rule, the bot, in its regex, check a code without any limit but all Discord invites have 8 characters. So, to avoid a member who has added a random char at the end of the link, by accident, take a sanction, I added a limit directly in all RegEx and in a substring for the part who works with the URL split.

### DIFF
--- a/backend/src/utils.test.ts
+++ b/backend/src/utils.test.ts
@@ -1,6 +1,6 @@
 import test from "ava";
 import z from "zod";
-import { convertDelayStringToMS, convertMSToDelayString, getUrlsInString, zAllowedMentions } from "./utils.js";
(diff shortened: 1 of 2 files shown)


iamshoXy:
Invites can have up to 10 characters and this would also break custom invite links

StanByes:
I didn't think this, I'll try to modulate it 
Maybe with an optional parameter in the match_invite rule to limit the char length ?
Any idea ?
Because, in our case, we have many member who take a sanction because they added, accidentally, a character at the end of or link

rubyowo:
That's pointless, you should just unwarn them manually, or just clean the message instead of warning entirely.

PR #479, opened 2024-06-01
fix: wrong key for counter's counter_trigger rule name

iamshoXy:
It appears that the trigger name is "triggers" instead of the actual AutoMod rule name.

Example path: ["counters", "nsfw_infractions", "triggers", "autoban", "condition"]

`ctx.path.length - 2` returns "triggers" while it should be `ctx.path.length -1` for the automod rule name.


iamshoXy:
I accidentally pushed my other changes to this branch ~ making a new PR 🙄

PR #480, opened 2024-06-01
fix: wrong key for counter's counter_trigger rule name

iamshoXy:
It appears that the trigger name is "triggers" instead of the actual AutoMod rule name.

**Example**: ["counters", "nsfw_infractions", "triggers", "autoban", "condition"]

ctx.path.length - 2 returns "triggers" while it should be ctx.path.length -1 for the automod rule name.

### DIFF
--- a/backend/src/plugins/Counters/types.ts
+++ b/backend/src/plugins/Counters/types.ts
@@ -32,7 +32,7 @@ export const zTrigger = z
       .optional(),
   })
   .transform((val, ctx) => {
-    const ruleName = String(ctx.path[ctx.path.length - 2]).trim();
+    const ruleName = String(ctx.path[ctx.path.length - 1]).trim();
 
     let reverseCondition = val.reverse_condition;
     if (!reverseCondition) {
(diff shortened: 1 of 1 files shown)


rubyowo:
LGTM

almeidx:
I don't quite understand the example in the description.

If `ctx.path` is `["counters", "nsfw_infractions", "triggers", "autoban", "condition"]`:
- `ctx.path.length - 2` will be `"autoban"` (the trigger name, configurable)
- `ctx.path.length - 1` will be `"conditions"` (constant property name)

Therefore I don't believe this change is correct

iamshoXy:
I have tested it and `ctx.path.length - 1` returned autoban in that case. I wrote my example manually and I think that `condition ` isn't actually in the array.

Dragory:
Good catch, thank you!

This was a copy-paste mistake since `- 2` is correct when transforming a nested `name` property, e.g. [here](https://github.com/ZeppelinBot/Zeppelin/blob/4ead472a2292ab527cfb2a31202d2395d9fb33da/backend/src/plugins/Automod/types.ts#L55).

PR #492, opened 2024-10-13
fix: build

almeidx:
The shared package doesn't have an index file


almeidx:
Resolved in f520c6f5d1c9616884351b0f9a74b3096750ff18

PR #500, opened 2024-11-09
fix: add updated mutes without a reason to the case log

rubyowo:
previously, when updating a mute, if you didn't add a reason, it wouldn't show up in the case logs (or case info).


Dragory:
Hey! Thanks for the PR. This was already fixed in https://github.com/ZeppelinBot/Zeppelin/commit/174e5cc23b8b3ac684f12c437ffaf88616774ac1 in the `next` branch, so I'm closing this one!

PR #501, opened 2024-11-09
feat: fix leap year rules, add year and month to the delay string

rubyowo:
this PR is essentially just https://github.com/EvanHahn/HumanizeDuration.js/pull/157 and adding `y` and `mo` to the delay string


Dragory:
Oh, looks like git didn't like that change in the merge base at all... I'll look at resolving those locally.

rubyowo:
moved to #503

PR #506, opened 2025-02-04
Update Client Events

xobradyn:
Changed client event "ready" and "error" to use the Events import as the "ready" will be replaced by "clientReady" in v15 of discord.js.

https://v15.discordjs.guide/additional-info/updating-from-v14.html#client


rubyowo:
This should be done alongside every other migration (there are a lot, ask me how I know)

PR #509, opened 2025-03-13
Port to Vue 3

TheKodeToad:
Vue 2 has been EOL for over a year

- Switched to `vue3-highlightjs` as `@highlightjs/vue-plugin` does not support v-highlightjs directive
- For some reason languages don't need to be registered any more and I have no idea why :trollface: 
- There are still some errors: image image
- Switching tabs in docs unhighlights the plugin. This seems to be a vue-router bug: https://github.com/vuejs/router/issues/2419.
- This is my first time working with vue so I would be happy for someone to take over :)

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -27,7 +27,7 @@
     "source-map-loader": "^4.0.1",
     "tailwindcss": "^1.9.6",
     "ts-loader": "^9.4.3",
-    "vue-loader": "^15.10.1",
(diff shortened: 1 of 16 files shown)


TheKodeToad:
I think it's in good enough a state.

Dragory:
@cla-bot check

Dragory:
This looks great, thank you for the PR! I'll merge this and also update the build process from Webpack to Vite.

PR #520, opened 2025-06-01
Fix loose matching false positives in MatchWordsTrigger

seeyebe:
## Problem
The current loose matching implementation creates too many false positives because it allows any characters between letters of filtered words. For example, filtering "hello" could incorrectly match "habcelldefgo".

## Solution
- Restricted loose matching to only whitespace and common separator characters (`[\\s\\-_.,!?]`)
- Fixed word boundary handling for loose matching patterns
- Now returns the actual matched text instead of empty string
- Added proper capture groups to identify which word triggered the match

## Testing
- [x] Normal word matching still works
- [x] Loose matching with various thresholds
- [x] Edge cases with punctuation
- [x] Verified reduction in false positives

### DIFF
--- a/backend/src/plugins/Automod/triggers/matchWords.ts
+++ b/backend/src/plugins/Automod/triggers/matchWords.ts
@@ -31,7 +31,6 @@ const configSchema = z.strictObject({
 
 export const MatchWordsTrigger = automodTrigger<MatchResultType>()({
   configSchema,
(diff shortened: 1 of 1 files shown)


Dragory:
@cla-bot check

Dragory:
This looks great, thank you for the PR!

PR #584, opened 2025-12-07
feat: resolve members on demand for mass actions

marawny:
we now actually ask discord if the member exists instead of just hoping

### DIFF
--- a/backend/src/plugins/ModActions/commands/massban/actualMassBanCmd.ts
+++ b/backend/src/plugins/ModActions/commands/massban/actualMassBanCmd.ts
@@ -11,7 +11,7 @@ import {
   isContextInteraction,
   sendContextResponse,
 } from "../../../../pluginUtils.js";
-import { DAYS, MINUTES, SECONDS, noop } from "../../../../utils.js";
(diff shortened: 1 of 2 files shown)


backend/src/plugins/ModActions/commands/massban/actualMassBanCmd.ts
```diff
@@ -45,7 +45,7 @@ export async function actualMassBanCmd(
 
   // Verify we can act on each of the users specified
   for (const userId of userIds) {
-    const member = pluginData.guild.members.cache.get(userId as Snowflake); // TODO: Get members on demand?
+    const member = await resolveMember(pluginData.client, pluginData.guild, userId);
```
almeidx: This is an inefficient way of fetching members. Given that the command executor can pass up to 100 members, worst case scenario this would make 100 API calls for fetching each member individually.

`guild.members.fetch()` accepts a `user` parameter which can receive an array of ids, and will use the gateway connection to retrieve all the users at once.

almeidx (changes requested):

PR #589, opened 2025-12-30
ensure codeblocks are closed

graphiteisaac:
This is really quite hard to test, but I think this should be able to fix the case reason codeblock problem we run into on the Overwatch server fairly often, eg:

image

If there's any proper testing I can do, or some way to clearly reproduce the issue, I'd be more than happy to try to do that and battle-test this a bit more.


graphiteisaac:
Have since signed and PRed the CLA.

zayKenyon:
> Have since signed and PRed the CLA.

Hi, your PR needs to point to ZeppelinBot/CLA:main, not your fork, then it can get merged and approved.

Dragory:
@cla-bot check

